PR #17, opened 2026-07-16
Fix Learn View/WebView2 pane corruption (stale mangled content + gray flashing)

jttdev:
Fixes the intermittent Learn View rendering defect documented in `notes/ABLETON-WINE-LEARNVIEW-FLICKER.md`. The initial hypothesis in that note (style-only mitigation) turned out to be necessary but not sufficient — there are two coupled defects, each fixed by one patch.

## Root cause 1 — patch 0035: the D3D target is never displayable

WebView2's `Intermediate D3D Window` carries `WS_EX_LAYERED | WS_EX_TRANSPARENT | WS_EX_NOREDIRECTIONBITMAP` and never receives layered attributes. On Windows, `WS_EX_NOREDIRECTIONBITMAP` bypasses layered redirection — DWM shows the DComp visual directly. Wine has no DWM: the dcomp stack emulates presentation with GDI blits into the window surface, but a never-attributed layered window is never displayed (wineserver marks it `SWP_NOREDRAW`), so every present/reblit landed in a surface that could not reach the screen — the pane showed Chromium's stale software-fallback sibling at old geometry (the mangled band).

**Fix:** strip the never-attributed `WS_EX_LAYERED` at swapchain adoption, restore on release (`__wine_dcomp_unlayered` prop). `WS_EX_TRANSPARENT` is kept so the software fallback still paints when Chromium abandons GPU compositing (the patch-0025 sidebar path is preserved). Also closes the patch-0030 black-hole window: WM_PAINT now paints a HALFTONE-scaled placeholder while the comp buffer is stale-gated instead of validating without painting.

## Root cause 2 — patch 0036: ancestors paint over the emulated visual

With the target displayable, a second defect surfaces: Live repaints its main view ~20x/s through a cache DC obtained **without** `DCX_CLIPCHILDREN` (SYSRGN probe shows the full client rect). Harmless on Windows — DWM composites the visual *above* the redirection surface. In Wine's shared-surface model those unclipped paints overwrite the pane (last-writer-wins), and once the reblit timer suspends 3 s after the last present, the pane overlap stays gray permanently. Measured live: pixel probes flip content→gray exactly at Live's damage-rect boundary, at the 200 ms reblit cadence.

**Fix:** emulate the DWM occlusion at both levels the host writes through — `get_visible_region` always subtracts visible descendants having `WS_EX_NOREDIRECTIONBITMAP` without `WS_EX_LAYERED` (exactly the adopted-target state), `get_surface_region` punches the same hole into the toplevel's window surface (the flush pushes the dirty *bounding* rect), and adoption nudges the toplevel with an async `FRAMECHANGED` SetWindowPos since the surface region is only recomputed in the surface owner's process.

## Scope guards / regression notes

- Siblings are deliberately unaffected: the software-fallback path for abandoned swapchains keeps painting (doc sidebar behavior preserved).
- JUCE DropShadower windows are layered **with** attributes → never match; SWAM-style d3d hwnd targets lack `WS_EX_NOREDIRECTIONBITMAP` → never match.
- Unadopted WebView2 shells keep stock semantics (the style strip only happens at dcomp adoption).

## Verification

Live 12.4.5b7, WoW64 container build, `WINEDEBUG=+dxgi,+dcomp` traces plus X-side probes: SYSRGN of every DC variant on Live's toplevel excludes exactly the pane rect after adoption; pane pixels stay on lesson content through reblit suspension and splitter resizes. Build audit passes with the two new entries (0035 has a binary fingerprint, `dxgi.dll` wide `__wine_dcomp_unlayered`).

## Known cosmetic residue (intentionally out of scope)

Because ancestors no longer paint under the visual, closing the pane leaves its rect on the last blitted frame until the host repaints it (mouse-over or a window resize heals it). We have a candidate follow-up for an explicit teardown reveal, but it needs more validation — kept out to keep this PR focused on the corruption fix. Happy to also upstream the SYSRGN diagnostic tool (`clipprobe.c`) used to prove root cause 2 if you want it in `tools/`.

### DIFF
--- a/patches/0035-dxgi-make-never-attributed-layered-dcomp-targets-dis.patch
+++ b/patches/0035-dxgi-make-never-attributed-layered-dcomp-targets-dis.patch
@@ -1,0 +1,208 @@
+From 0b6ee3cfeaf545e2c270aa1ca3d132bae5aa778a Mon Sep 17 00:00:00 2001
+From: dist <build@localhost>
+Date: Wed, 15 Jul 2026 09:52:27 +0200
+Subject: [PATCH] dxgi: make never-attributed layered dcomp targets displayable
+MIME-Version: 1.0
+Content-Type: text/plain; charset=UTF-8
+Content-Transfer-Encoding: 8bit
+
+On Windows a WS_EX_NOREDIRECTIONBITMAP DComp target window bypasses the
+layered redirection surface entirely: WS_EX_LAYERED is set but never given
+attributes, and DWM displays the composition visual directly.  Wine has no
+DWM; this dcomp stack emulates presentation with GDI blits into the window
+surface — but Wine also faithfully emulates 'layered without attributes is
+not displayed', so those blits never reach the screen.  WebView2's Learn
+View pane then shows the sibling Chrome_RenderWidgetHostHWND's stale
+software-fallback frame, drawn once at old geometry: the mangled band and
+flicker (XDamage-verified: zero X-side damage from correct-size blits;
+stripping WS_EX_LAYERED live healed the pane instantly).
+
+Three coupled fixes:
+- WM_WINE_DCOMP_SET_TARGET: strip never-attributed WS_EX_LAYERED from the
+  adopted target (both full and popup mode); restore it if the swapchain is
+  released while the window survives.  WS_EX_TRANSPARENT stays, so the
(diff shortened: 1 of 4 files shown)


shibco:
Thanks, @jttdev - this PR was a good starting point but needed some additional work. I won't merge this, but the next version will have fixes that this PR pointed me towards.

Issue #16, opened 2026-07-16
NixOS + NVIDIA (via steam-run): Direct2D/EGL renderer fails with `driver (null)`, Live crashes — plus a niri window-mapping win


xenicle:
First, thanks for this project — the approach is genuinely nice, and I hit one clear blocker plus one positive finding worth sharing.

## Environment

- **Distro:** NixOS (no native FHS `/usr`, so the installer + runtime are run through `steam-run`, which provides an FHS sandbox; glibc 2.42)
- **Compositor:** niri (Wayland) with `xwayland-satellite`
- **GPU:** NVIDIA RTX 2080 Ti, proprietary driver 595.71.05
- **Ableton:** Live 12 **Standard**
- **shibco:** `install-ableton-latest.run` v2026.07.14.2 → `wine-d2d1-nspa-11.11`
- Ran as: `steam-run bash install-ableton-latest.run` (with `cabextract` and `zstd` added to PATH). Prefix + runtime install and the Ableton install both completed fine.

## ✅ What works (a positive note for niri users)

The patched Wine **maps Ableton's main window on niri** — where a plain/host Wine does **not** (host Wine only maps simple windows like `winecfg` there, not Live's main window). So your Wine patches appear to fix Live's window mapping on the wlroots/`xwayland-satellite` family. Nice result — niri isn't in your supported list, but it works for the display side.

## ❌ The blocker: Direct2D/EGL can't reach the NVIDIA driver in the FHS → crash

Shortly after launch, Live becomes unstable / crashes. The log is flooded with:

```
libEGL warning: pci id for fd NN: 10de:1e07, driver (null)
libEGL warning: egl: failed to create dri2 screen
```

`10de:1e07` is the NVIDIA GPU. Inside the `steam-run` FHS, **mesa's `libEGL` is loaded and can't drive the NVIDIA GPU**, so the Direct2D/D3D render path fails in a tight loop. This lines up with the launcher's own comment ("*fixme spam from the D3D poll loop stalls Live's UI thread*") — here the loop is failing EGL rather than just spamming, and it destabilizes the UI thread.

## What I tried (didn't resolve it)

Exposing NixOS's NVIDIA userspace driver into the FHS:

```sh
export LD_LIBRARY_PATH=/run/opengl-driver/lib:$LD_LIBRARY_PATH
export __EGL_VENDOR_LIBRARY_DIRS=/run/opengl-driver/share/glvnd/egl_vendor.d
[...]

PR #22, opened 2026-07-18
Add native fractional DPI support for Hyprland

astrazds:
## Summary

- add a pure, testable DPI policy resolver;
- detect Hyprland with `xwayland:force_zero_scaling` enabled;
- use exact application-side DPI (`round(96 * scale)`) without an IFEO override;
- apply the same policy during prefix setup and every launcher start; and
- retain the existing 100% and legacy 192-DPI fractional behavior for other desktops.

Unknown compositor/XWayland strategies continue to preserve existing prefix values.

## Why

At Hyprland scale `1.666667`, `detect-scale.sh` normalizes the scale to `1.67`, but the existing launcher treats it as uncalibrated and leaves a fresh prefix at 96 DPI. With `xwayland:force_zero_scaling=true`, Live should scale itself at the monitor's native logical DPI:

```text
round(96 * 1.666667) = 160
```

Testing also showed that Live becomes per-monitor aware without an executable-specific `dpiAwareness` override, so the native policy removes a stale IFEO value rather than adding one.

## Test system

- Arch Linux / Omarchy
- Hyprland 0.55.4
- `xwayland:force_zero_scaling=true`
- 3840x2160 at 159.75 Hz
- compositor scale `1.666667` (`detect-scale.sh`: `1.67`)
- NVIDIA GeForce RTX 5080, driver 610.43.03
- Wine runtime 11.11 from release `v2026.07.17.3`
- Ableton Live 12 Suite 12.4.3

## DPI candidate results

| LogPixels | IFEO | Result |
| ---: | ---: | --- |
| 96 | absent | Functional and coordinate-correct, but visibly underscaled |
| 160 | 2 | Crisp, correctly sized, and passed geometry/hit-test/manual checks |
| 160 | absent | Same DPI awareness and geometry; passed; selected policy |
| 192 | 2 | Functional, but approximately 20% oversized for a 1.666667 display |

For the selected candidate, Live and Learn View report 160 DPI. The tiled main window measured 1138x1278 logical and 1898x2130 physical, matching the compositor scale within rounding tolerance. Main-window and embedded-child hit tests resolved correctly.

The modified launcher was then tested end to end from a cloned 96-DPI/no-IFEO prefix. It automatically selected `native:160`, wrote `LogPixels=160`, kept IFEO absent, and reproduced the passing geometry and manual behavior.

## Validation

- `tests/test-dpi-policy.sh`: PASS for Hyprland 1.00, 1.25, 1.50, 1.666667/1.67, 1.75, and 2.00;
[...]

PR #23, opened 2026-07-18
Select Wine GLX for active NVIDIA displays

astrazds:
## Summary

- detect an active NVIDIA display or explicit PRIME offload and select the Wine GLX backend
- preserve the existing backend for Intel, AMD, and unknown systems
- add `ABLETON_OPENGL_BACKEND=auto|preserve|egl|glx` as an escape hatch
- ship the policy through both repository and self-extracting installer paths
- add focused policy regression tests

## Why

On the tested RTX 5080 / NVIDIA 610.43.03 Wayland + XWayland system, Wine EGL exposes RGBA8 configs on a 32-bit ARGB visual. Wine intentionally does not advertise those as ordinary window formats when their visual depth differs from the X11 default, so Ableton BaseView receives zero matching window-capable pixel formats.

Wine GLX exposes compatible RGBA8 and sRGB formats on the same system. Selecting GLX resolves the failure without weakening the Wine visual-depth safety check or forcing a software renderer.

This is related to #16, but does not claim to fix its separate `steam-run` / NixOS library-selection problem.

## Validation

- real Ableton Live 12 launch through the modified launcher
- confirmed `libGLX_nvidia.so` and `libnvidia-glcore.so` loaded
- confirmed Live registered as a graphics process in `nvidia-smi pmon`
- `glchild.exe`: 8 formats; OpenGL 3.2, 3.3, and 4.1 core contexts pass
- BaseView sRGB and non-sRGB pixel-format selections pass
- empty probe stderr; no Mesa `driver (null)` warnings
- policy test suite passes all 8 cases
- shell syntax and `git diff --check` pass
- runtime build audit passes all 54 checks
- self-extracting installer builds, verifies, and contains the new policy file

## Scope

Automatic behavior changes only when NVIDIA is actively driving a display or PRIME render offload is explicitly enabled. Existing Wine backend configuration is preserved everywhere else.

### DIFF
--- a/README.md
+++ b/README.md
@@ -130,6 +130,7 @@ Mostly unnecessary. But in case you need them:
 - `ABLETON_WINEPREFIX` prefix path (default `~/.wine-ableton`)
 - `ABLETON_DPI_MODE` `auto` | `preserve` | `100` | `fractional`
 - `ABLETON_THEME_MODE` `auto` | `dark` | `light` | `preserve` — the launcher syncs Live's light/dark theme key to the desktop scheme on every start; this overrides it
+- `ABLETON_OPENGL_BACKEND` `auto` | `preserve` | `egl` | `glx` — `auto` selects GLX when an NVIDIA GPU is actively driving a display, avoiding NVIDIA EGL/X11 pixel-format incompatibilities; other GPUs keep their existing backend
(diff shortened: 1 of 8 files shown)


shibco:
This is fixed in a similar way to #22! Thank you again!! [Stability update (2026.07.18.1)](https://github.com/shibco/ableton-linux/commit/c2c468191b10d4340dbd8581f4c0234acb2041dd)

PR #25, opened 2026-07-18
feat: added app & mime icons

yioannides:
As we briefly discussed in Discord, I created (Adwaita-styled) app & mime icons for Ableton Live.

Few things to note:

- The `application-ableton-live.xml` mime declaration file needs to be relocated in `$HOME/.local/share/mime/packages` and then `update-mime-database $HOME/.local/share/mime/`, let's include that in the installation.

- I'm not sure how to source the MIME icons. They either need to be on the user's current theme (hoping it's a custom theme that already sources `scalable/mimetypes/` in its `index.theme` file) or the default `hicolor` theme whose local path does not have an `index.theme` file. If you have any ideas on how to go about it, please let me know. (technically speaking, app icons can be downloaded and manually sourced like `Icon=@HOME@/.local/share/wine-ableton/icons/scalable/apps/live-xxxx.svg`, but by doing so, they will not automatically inherit their companion symbolic icons)

- Since there are multiple versions of Ableton Live, should we instead generate `.desktop` files within the installation script based on the package name being installed? For example, the `Name`, `Path` and `Icon` variables will be unique (also make sure to also source the `MimeType` variable):

```
[Desktop Entry]
Name=Ableton Live 12 Suite
Comment=Music production (Wine, patched d2d1-dcomp stack)
Exec=/home/yiannis/.local/bin/ableton-live
Type=Application
StartupNotify=true
Path=/home/yiannis/.wine-ableton/drive_c/ProgramData/Ableton/Live 12 Suite/Program
Icon=live-suite.svg
MimeType=application/ableton-live;
StartupWMClass=ableton live 12 suite.exe
Categories=AudioVideo;Audio;
```


yioannides:
I'm thinking about closing and redirecting this PR to [Morewaita](https://github.com/somepaulo/MoreWaita): assigning MIME icons from a non-icon pack is way too hacky (and potentially pervasive), plus the design is already Adwaita-ready (might look weird on a KDE desktop), so I might as well...

PR #27, opened 2026-07-18
Initial Flake Setup

Version33:
This adds a Nix Flake for setting up and running ableton-linux.

   `nix run .#setup-prefix` — creates the Wine prefix at ~/.wine-ableton (fonts,
   VC++ runtime, PipeASIO, DPI/theme) and installs Ableton from the newest
   ableton_live*.zip found in $ABLETON_INSTALLER_DIR (default ~/Proprietary).
   No zip there → prefix is still created and the manual install commands are
   printed; rerun after adding it. The winetricks payload cache is vendored into
   the package, so the Live 12 recipe (corefonts, vcrun2022, mfc42) installs
   offline; the Live 11 recipe still downloads vcrun2019/gdiplus.

   `ABLETON_LIVE_AUTOINSTALL=1` opts in to running the installer silently,
   deferring its UI including EULA acceptance to Live's first launch.

   `nix run .` — launches Ableton Live (errors with instructions if setup-prefix
   hasn't installed it yet).

   Desktop entries are rendered into `share/applications/` at build time, so a
   profile install or `environment.systemPackages` registers them automatically.
   PipeASIO config.ini keys can be pinned declaratively via
   `ableton-wine.override { pipeasioSettings = { ... }; }`; `PIPEASIO_*`
   environment variables still override per launch.

   **Changes shared with the tarball path** (not flake-only):

   - `setup-run-header.sh`: installer downloads are version-sorted by basename
     (a directory name can no longer outrank the version when zips sit in both
     the stick dir and ~/Proprietary), and the chosen download's major now pins
     the prefix recipe — a Live 11 zip gets the Live 11 winetricks recipe instead
     of landing in a Live 12 prefix. An explicit ABLETON_LIVE_VERSION still wins.
   - `setup-prefix.sh`: the bundled winetricks cache is consumed through per-verb
     symlinks in a writable tmpdir, so it also works from a read-only location
     (the Nix store); verbs missing from the cache download exactly as before.

   **Disclaimer**: This was coded with an AI, Claude Fable in this case. It does
   works though, just needs more testing on more machines to make sure things
   were done right.


shibco:
Heads up, I don't know if this will affect your work, but `2026.07.18.1` is dropping and has a massive backlog of changes - please let me know if something breaks or if you have any questions about it.

Version33:
> Heads up, I don't know if this will affect your work, but `2026.07.18.1` is dropping and has a massive backlog of changes - please let me know if something breaks or if you have any questions about it.
> 

ideally this whole setup will be able to sit on top of anything that changes  
but I guess we shall see

Version33:
that should do it  
would definitely like some review

shibco:
Thanks for this. The Nix work is well researched. The derivation faithfully ports container-build.sh, the SERIES.sha256 patch verification is stricter than the container build, the sed anchors fail loud, and the README claims all check out against the code. There are a couple of fixes first:                                                                                                                                                  

1. The default package likely does not build. `nix/wine.nix:198` deletes the Wine import libraries (`rm -f $out/lib/wine/*-windows/*.a`) in postInstall, but `nix/pipeasio.nix:43-46` links against exactly those files (`-lodbc32 -lole32 -luuid -lwinmm -luser32`). The container build works because it builds PipeASIO at step 4 and prunes at step 5. From what I can see, the flake inverts that order.
2. The `setup-prefix.sh` includes a conflict that reverts a recently-pushed issue #29 fix. (sorry!) Main's `2026.07.19.1` from today replaced `seed_options_txt` with `strip_options_txt (6b4e008)`. Your branch still seeds `-DontCombineAPCs` which was determined to create PipeASIO issues.                                                                                                                 
3. Unless I'm mistaken, Live 11 installs into a Live 12 prefix. Unpinned runs default the winetricks recipe to 12 (`setup-prefix.sh:323`), but step 6 installs whatever zip it finds (`:595`). A Live 11 zip in `~/Proprietary` misses `vcrun2019 gdiplus`. Detect the version from the zip or pin consistently.                                                                             
4. Step 6 runs any `~/Proprietary/ableton_live*.zip` with `/VERYSILENT` by default. That defers EULA acceptance, contradicts the script's own header comment, and turns a prefix-refresh tool into something that executes whatever sits in a well-known directory. Convenient but also kind of dangerous. This also affects tarball users, which the PR description does not mention. Make it opt-in (`ABLETON_LIVE_AUTOINSTALL=1`) or interactive by default, and disclose the change in the PR body.                                                                                                         
   
Overall, this is pretty good! Thank you! I think it's pretty close to being merged.

Version33:
I'll get to work on that!

Version33:
removed and added nix config things  
it's kinda wide, but I would very much like to config via nix instead of the config.ini file

shibco:
Thank you for this work, and nice job. All four items from my last review are resolved.

1. Import-lib order now matches the container: `nix/wine.nix:184-189` keeps the `*.a` libs, pipeasio links against the unpruned output, `nix/ableton-wine.nix:101-105` prunes last
2. `strip_options_txt` survives byte-identical; the upstream merges reverted nothing.
3. Zip and recipe major resolve together up front (`setup-prefix.sh:311-350`).
4. Auto-install is opt-in, the `.run` pins `ABLETON_LIVE_AUTOINSTALL=0`, EULA deferral disclosed.          

Two fixes before merge, and some nice to haves:

1. **Revert `81f6632`, `PIPEASIO_*` env vars do exist in 1.2.2.** The PE half of the driver reads ten of them via `GetEnvironmentVariableA` (`src/asio.c:1875-1947`, "Environment variables override INI values"); the driver's own README documents them. A `getenv` grep misses them because the Windows side uses the Win32 API. The old documation here was right. But, as written this deletes the `PIPEASIO_PREFERRED_BUFFERSIZE=512` crackle workaround and regresses the tarball path.
2.**`setup-run-header.sh:91-92`, now, `sort -V` runs over full paths from two directories**, so the directory name wins over the version. Zips in both the stick dir and `~/Proprietary` can pick the older Live. This is the Live-11-into-a-12-recipe bug again, through the `.run`. Instead, I'd suggest sorting this by basename and/or deriving the version from the chosen zip too.

Nice to haves: 
- the Nix package ships winetricks without the vendored cache, so `nix run .#setup-prefix` needs network access (README should say so)
- `nix/wine.nix` drops the container's `winealsa.so` presence gate;
- `check-ntsync.sh` ships without its probe; 
- desktop templates still have the classic `@HOME@` placeholders; 
- Live 11 Max fixup needs the `nix run .#setup-prefix -- --post-first-run` invocation documented.              

Thanks again for this. We are very close!

Version33:
another pass done

Version33:
info from claude about latest push:
```
Rebased onto 2026.07.29.1. Adapted the flake to the 11.13 base — wineSrc follows wine-base-5c23dd1c.tar.zst and the version references move to 11.13.      
                                                                                                                                                         
Added nix/ableton-linkd.nix so the packaged runtime carries the Link anchor from #80/#68: both launchers, setup-link.sh and the unit's ExecStart resolve   
ableton-linkd under $HOME, which a store install never populates, so the flake stages it and repoints those defaults. ABLETON_LINKD still overrides. Built 
the daemon from the vendored SDK the way tools/build_ableton-linkd.sh does, and gated it the way install.sh does — it must run and link nothing beyond the 
host C runtime.                                                                                                                                            
                                                                                                                                                            
Also documented --post-first-run, the last open item from your review.                                                                                     
                                                                                                                                                            
Verified: nix flake check; full nix build with the ntsync, winealsa and PipeASIO-registration gates passing; setup-prefix end to end in a throwaway prefix 
(exit 0, vendored winetricks cache, no network); the Live 11 Max fixup invocation both with and without a prefix.
```

Immabed:
One consideration for adding the flake is potentially adding a releases branch to the repo and recommending that as the flake url for nix users, otherwise users of the flake will always get the most recent commit to `main` when running `nix flake update` (unless they are pinning their config to a specific commit or against a cloned repo).

Eg. If there was a `release` branch that was only updated for every ableton-linux release, nix users could use the following config to follow the releases.

`ableton-linux.url = "github:shibco/ableton-linux/release";`

realitymolder:
> One consideration for adding the flake is potentially adding a releases branch to the repo and recommending that as the flake url for nix users, otherwise users of the flake will always get the most recent commit to `main` when running `nix flake update` (unless they are pinning their config to a specific commit or against a cloned repo).
> 
> Eg. If there was a `release` branch that was only updated for every ableton-linux release, nix users could use the following config to follow the releases.
> 
> `ableton-linux.url = "github:shibco/ableton-linux/release";`

you refer to beta/release or more of ableton12/ableton11 branches? or both?

Immabed:
> you refer to beta/release or more of ableton12/ableton11 branches? or both?

Releases of ableton-linux. Live 11 and 12 are both supported by the same project so no need for separate branches.

Version33:
> One consideration for adding the flake is potentially adding a releases branch to the repo and recommending that as the flake url for nix users, otherwise users of the flake will always get the most recent commit to `main` when running `nix flake update` (unless they are pinning their config to a specific commit or against a cloned repo).
> 
> Eg. If there was a `release` branch that was only updated for every ableton-linux release, nix users could use the following config to follow the releases.
> 
> `ableton-linux.url = "github:shibco/ableton-linux/release";`
> 
> 

That would be for a maintainer to do, but nix can also just pin to a specific commit should someone want to avoid updating for some reason 
(there's also rolling back your config of you use NixOS for when something breaks)
if @shibco wants to do "stable releases" for something like that it would require no extra work on this PR
but at the current stage of the project, I don't think that will be of much benefit

Immabed:
The main potential problems are things breaking for nix users, and a lack of consistency between nix and other installation methods, since nix users won't usually be on the specific commit that aligns with an ableton-linux release, which could make bug reporting more confusing as well.

Certainly not necessary, and easy for a nix user to pin a specific commit (I've been doing that), but worth considering I think, because the distribution method for the flake is so different from the regular releases.

Definitely not a gating factor for this PR.

Immabed:
Minor issue with the nix documentation regarding ntsync.

README.md says
> Your host needs a running PipeWire daemon and /dev/ntsync (kernel 6.14 or newer with the ntsync module; scripts/check-ntsync.sh verifies this).

But `check-ntsync.sh` specifically looks for the wine root at `$HOME/.local/opt/wine-d2d1-nspa-11.13`, which will not be the location on nix systems (it'll be somewhere in `/nix/store/`). Running the script when using the flake results in a `!! no wine at ...` message.

Manually updating the wine root in the script with the proper path worked fine.

Simply not recommending the script as a check in the Nix section of the documentation is probably sufficient, otherwise you're needing to find the wine root on nix systems.

However, recommendations in the 'Instruments and Effects' section include commands referring to the wine root at `$HOME/.local/opt/wine-d2d1-nspa-11.13/bin/wine` which also won't work for users of the nix flake, since there is no wine at that location.

shibco:
Good news! this is getting close.
To everyone who worked on this, I appreciate you keeping up with a moving target. This PR has not been forgotten!

Some thoughts:

1. The nix/wine.ix  build omits GStreamer and its plugins and does not fail with that omission.  The full build succeeds, but its output contains no winegstreamer.so. The absence of winegstreamer will break MP3, MP4 and WMA importing. 
2. The required Bitstream Vera fonts are absent, and these are essential to prevent Max4Live devices that have missing fonts from freezing the live interface. This is a Max4Live issue, but unlikely to be seen in a Windows or macOS build. This is essential, and the build should fail without these faces.
3. User configuration records disposable Nix store paths and the desktop-handler installation follows a similar pattern. This is unsafe with the README’s recommended unrooted nix run workflow: unreachable store objects may be garbage-collected, and upgrades leave copied services pointing at an older package hash.
4.  Last thing, I noticed that the  “Newest installer” selection is incorrect across Ableton editions in `setup-prefix.sh` Because the edition precedes the version, lexical edition ordering can dominate version ordering.

shibco (changes requested):
See comment above for changes.

Version33:
plugged feedback into claude
spat out this after..... a long while

some claude notes from that
```
All four fixes are in, each its own commit, and the full flake builds green from cold.                                                                       
                                                                                                                                                              
 Commits                                                                                                                              
                                                                                                                                                              
 ┌─────────┬─────────────────────────────────────────────────────────────────────┐                                                                            
 │         │                                                                     │                                                                            
 ├─────────┼─────────────────────────────────────────────────────────────────────┤                                                                            
 │ 154ad42 │ nix: build winegstreamer, and fail the build without it             │                                                                            
 ├─────────┼─────────────────────────────────────────────────────────────────────┤                                                                            
 │ f6edd37 │ nix: ship the Bitstream Vera faces, and fail the build without them │                                                                            
 ├─────────┼─────────────────────────────────────────────────────────────────────┤                                                                            
 │ 48a245b │ keep store paths out of user configuration                          │                                                                            
 ├─────────┼─────────────────────────────────────────────────────────────────────┤                                                                            
 │ 4adda66 │ setup-prefix: pick the newest installer by version, not by edition  │                                                                            
 └─────────┴─────────────────────────────────────────────────────────────────────┘                                                                            
                                                                                                                                                              
 1 — GStreamer. wine's configure treats missing gstreamer dev files as a notice, so the flake silently dropped winegstreamer while succeeding. Added          
 gstreamer + gst-plugins-base (link deps) and good/bad/ugly/libav (the actual decoders — the .run borrows codecs from the host, a minimal NixOS has none),    
 --with-gstreamer so a missing dep is a configure error, plus postInstall gates on winegstreamer.so/.dll, its libgstreamer-1.0.so.0 DT_NEEDED, RUNPATH        
 self-sufficiency, and the mpg123/isomp4/asf/libav plugins. bin/wine exports GST_PLUGIN_SYSTEM_PATH_1_0. Verified on the shipped package: winegstreamer.dll   
 loads as builtin in both WoW64 halves, all four decoder families on the path.                                                                                
                                                                                                                                                              
 2 — Bitstream Vera. Staged into share/ableton-wine/vendor/fonts where setup-prefix.sh's kit_root() looks, with COPYRIGHT.TXT alongside; installPhase         
 verifies vendor/bitstream-vera.sha256 (pins content and completeness) and installCheckPhase asserts all ten faces at the resolved path — the build now fails 
 rather than shipping a runtime that can freeze Live. scripts/check-m4l-fonts.sh: 7/7 offline+assumption checks pass.                                         
                                                                                                                                                              
 3 — Store paths in user config. New scripts/runtime-link.sh maintains ~/.local/share/ableton-wine/runtime and, on a store root, registers it as an indirect  
 GC root (same mechanism as nix build's result; no privileges). Verified against the real store package: link → the store path, rooted, handler Exec= and the 
 unit's ExecStart= both via the link with zero store paths recorded, and an "upgrade" re-points the link with the installed files untouched. Entries that     
 already name a store path, or whose absolute Exec is gone, now count as needing repair; hand-edited relative ones still stay. Tarball flow unchanged (%h/... 
 preserved verbatim).                                                                                                                                         
                                                                                                                                                              
 4 — Installer selection. sort -V on the name ranked edition first: with a real fixture set it picked ableton_live_trial_11.3.35 over every 12.x. Now sorts   
 on a version key from the basename — picks 12.10.0 over 12.5.1, and the ABLETON_LIVE_VERSION=11 pin still selects the 11.                                    
                                                                                                                                                              
 One thing worth flagging                                                                                                                                     
                                                                                                                                                              
 Fix 1 initially made wine's pre-existing smoke gate SIGSEGV 100% of the time — and it cost real effort to pin down, because the artifact was never at fault  
 (the same tree ran 20/20 in any other sandbox and on the host). Nix runs builders with ASLR disabled, so the layout is a deterministic function of the       
 environment; adding buildInputs shifted it into a collision with wine's preloader reservation, before ntdll can print anything. Every build where I edited   
 the phase passed, which is what disguised it as flakiness. The gate now runs wine under env -i with a fixed environment, so it tests wine instead of the     
 build environment — cold builds green 3×. Same latent fragility exists in ableton-wine's PipeASIO gate; it passes today, and I left it alone rather than     
 churn a working gate.
 ```

**Comment by ClickSentinel** [comment]
Re-reviewed at `8508460`. Read the diff and ran the shell parts; no nix here, so I didn't build the flake. Merge against main is clean now.

All four items from your last round are properly fixed, and I tested them rather than just reading. One thing came in with the merge, though.

### Blocking

**The merge lost main's bound on the boot wait.** Main had `timeout 30 "$WINESERVER" -w` at `setup-prefix.sh:313-320`, falling back to `wineserver -k` and a second bounded wait on timeout. `8508460` replaces that with a bare `settle` at `:427`, which has no timeout and swallows errors. The tray-agent scrub survived the merge; the bound didn't, and main's comment explaining why it was there went with it.

That comment is the argument: the scrub matches known image names, and it says itself that those names vary by driver generation. The bound was the backstop for the agent it misses. Disabling `MicrosoftEdgeUpdate.exe` up front covers the other cause, not this one.

Fix: keep `settle` for the narration, but give it a timeout and reuse main's rc-124 branch.

### Should fix

**`ableton_runtime_link` re-points the link non-atomically.** It does `rm -f` then `ln -sfn`, which isn't a rename, and `48a245b` now has the desktop handlers and the linkd unit running through that path. Twelve concurrent launchers with a reader polling `$link/bin/ableton-live`: same root, 0 misses in 4000 reads; two different roots, 1490. The clean arm passes because the "already current" check returns before the `rm`, not because the write is safe.

You need two runtimes launching at once to hit it (a beta beside a stable, say), so it's not the common case. Fix: `mv -T` from a temp name, and on the store branch skip the `rm` when the link already points into the store — by your own comment that's the case `nix-store` can overwrite anyway.

**pipeasio patches are still applied unverified.** `nix/wine.nix` checks the wine half against `SERIES.sha256`, `nix/pipeasio.nix` doesn't check its half at all, and BUILDING.md claims the check without qualifying it. Fix: same `sha256sum -c` on the `pipeasio/` prefix, or reword.

**Still no `ABLETON-WINE-PATCH-STACK.txt`, and `build-audit.sh` never runs.** A Nix-built runtime can't be audited or identified in a bug report. Its `must` list is also what would have caught the winegstreamer and font gaps to begin with — worth wiring in rather than reimplementing gates one at a time. Fix: stamp it and run `build-audit.sh $out`.

### Minor

- The flake now pins its own decoders (`good`/`bad`/`ugly`/`libav`) instead of borrowing from the host like the `.run` does. Sensible for NixOS, but it's a real difference between install paths and it pulls ffmpeg into the closure — worth a line in the README.
- `runtime-link.sh` trips SC2148. You added `# shellcheck shell=bash` to `detect-theme.sh` for the same reason.
- The smoke-gate SIGSEGV deserves a comment in the file, not just the thread. Same for the PipeASIO gate you left alone — agreed that's the right call, but nobody should have to rediscover it.
- Desktop entries still hardcode `ableton live 12 suite.exe`; `flake.lock` still on 2026-07-15.

### Checked

Installer selection, old against new, four fixtures: the old code picked `suite_12.0.1` over `intro_12.4.3` and an 11.3.42 trial over both. New code gets all four right, including 12.10 over 12.5. Fonts: manifest verifies, the ten staged faces match what `install_maxplug_fallback_fonts` looks for, staged where `kit_root()` resolves — and asserting `vendor/winetricks` is still there is a nice way to pin the assumption underneath it. winegstreamer: `--with-gstreamer` turns the silent drop into a configure error, and the plugin-path and RUNPATH gates go further than I'd have asked for. Registration after the merge: 81 patches, manifest matches, BASE.txt count right. `bash -n` and shellcheck clean apart from the above.

**Comment by Immabed** [comment]
This seems like a nix specific issue, related to updating the .desktop files.

I was scratching my head because it seemed like issue #100 was recurring for me, but I've nailed it down to the .desktop files not updating, launching via the terminal command seems to use the latest build.

When launching ableton-live via a terminal, I get the following errors:
```
ableton-live: reinstating wine-protocol-ableton.desktop (it names a garbage-collectable store path)
/nix/store/wx0ycxzr4gvy0swrqvbi6kjcdvpa6w61-ableton-wine-11.13/libexec/ableton-live: line 135: /home/immabed/.local/share/applications/wine-protocol-ableton.desktop: Permission denied
ableton-live: reinstating wine-extension-auz.desktop (it names a garbage-collectable store path)
/nix/store/wx0ycxzr4gvy0swrqvbi6kjcdvpa6w61-ableton-wine-11.13/libexec/ableton-live: line 135: /home/immabed/.local/share/applications/wine-extension-auz.desktop: Permission denied
```

Checking the .desktop files there are referring to older builds of ableton-live, so is using older builds of the custom wine. I'm not sure if this was ever working to be honest.

I have the flake installed in my nixos config, fwiw.

realitymolder:
anything else left?

Version33:
general fixes removed in https://github.com/Version33/ableton-linux/pull/1
moved to https://github.com/shibco/ableton-linux/pull/212
so this should only contain flake code

realitymolder:
Seems that there was an issue with the script, I've sent a fix for that in the #212 general fixes.

Username404-59:
Btw could you add [flake-compat](https://github.com/NixOS/flake-compat) for people who are not using flakes?

Version33:
> Btw could you add [flake-compat](https://github.com/NixOS/flake-compat) for people who are not using flakes?

I don't see a need to; pretty sure `flake-compat` can be used with a flake that doesn't have it already so it's a little superfluous to add it when majority of systems that use nix also use flakes.  
Although, I could be wrong here; maybe there are a lot of people who don't want to enable flakes.

realitymolder:
these days flakes are pretty much de-facto. BUT! even if we will add a feature like that, let us first add the flake to the main branch. We can always add another feature.

Username404-59:
> I don't see a need to; pretty sure flake-compat can be used with a flake that doesn't have it already

It's not meant to be used like that AFAIK

>  it's a little superfluous to add it when majority of systems that use nix also use flakes.

> these days flakes are pretty much de-facto. BUT! even if we will add a feature like that, let us first add the flake to the main branch. We can always add another feature.

I mean some people do avoid enabling flakes & it's just a few lines to copy-paste which is why I mentioned it; entirely up to you

Username404-59:
Hello again, I found an issue

Somewhere it does `ln -s ${lib.getBin unzip}/bin/unzip  $out/bin/unzip` which makes nixos-rebuild fail on my system because I already have unzip in my packages
No idea what the AI was thinking but shouldn't `unzip` it be in `buildInputs` instead?

Version33:
> Hello again, I found an issue
> 
> Somewhere it does `ln -s ${lib.getBin unzip}/bin/unzip  $out/bin/unzip` which makes nixos-rebuild fail on my system because I already have unzip in my packages
> No idea what the AI was thinking but shouldn't `unzip` it be in `buildInputs` instead?

will look into this when I'm home  
though I have unzip on my system as well and haven't seen this issue  
will let you know

Username404-59:
> _bro deleted their comment, I'll take that as an "issue resolved_"

<sup><sub>what</sub></sup>

Version33:
> > _bro deleted their comment, I'll take that as an "issue resolved_"
> 
> <sup><sub>what</sub></sup>

mobile app bugged out, dw about it  
this is a bad place to be taking  
consider joining the discord :)

shibco:
bruh, we merged it #237!

PR #37, opened 2026-07-20
Register PipeASIO .dll silently

jackson-57:
`regsvr32` has an optional argument `/s` which prints the results of the command to the console instead of in a dialog box.


shibco:
This is a no brainer, lol. Thank you!

PR #54, opened 2026-07-22
detect-scale: add a COSMIC probe

ClickSentinel:
## Summary

COSMIC (System76's compositor) matched none of the existing display-scale probes (GNOME/mutter DBus, KDE's kscreen-doctor, sway, Hyprland) or the Xft.dpi fallback, so auto-detection always failed on it: harmless on `--refresh` (falls back to preserving the existing prefix's DPI block), but a hard failure on a fresh prefix, which then requires an explicit `ABLETON_DPI_MODE` to proceed at all.

Adds `_ads_cosmic`, following the shape of the existing probes. `cosmic-randr list` reports per-output `Scale` and an `Xwayland primary` flag identifying the output Ableton actually renders on.

## Testing

Tested live against a real COSMIC session (3 monitors, `XDG_CURRENT_DESKTOP=COSMIC`) at both 100% and 125% scale — scale detected and mapped to the correct DPI block end to end.

Bucketed with the generic (non-GNOME) scaling family, confirmed rather than assumed: at 125% scale, `xrandr` on the Xwayland display reports the monitor's native mode (2560x1440) unscaled, not an integer-upscaled framebuffer (which would show ~3200x1800). COSMIC hands XWayland the native resolution and expects application-side scaling — the same model as sway/Hyprland/KDE, not mutter's.

### DIFF
--- a/scripts/detect-scale.sh
+++ b/scripts/detect-scale.sh
@@ -69,13 +69,9 @@ _ads_cosmic() {
     local out prim
     out="$(timeout 5 cosmic-randr list 2>/dev/null | sed 's/\x1b\[[0-9;]*m//g')"
     [ -n "$out" ] || return 1
-    # One "<output> (enabled|disabled)" block per monitor. A disabled output (e.g. a
-    # closed laptop lid with an external monitor attached) still reports a Scale on
-    # newer cosmic-randr, so disabled blocks are excluded entirely rather than just
-    # deprioritized -- their scale must never be picked, including as the fallback.
-    # The enabled block with "Xwayland primary: true" is the one Ableton actually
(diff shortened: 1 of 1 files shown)


shibco:
Woo! Cosmic support!!
One fix before merge. cosmic-randr prints `Scale:` for disabled outputs too, and older COSMIC doesn't print `Xwayland primary:` at all. So we need to add something to ensure that the fallback doesn't detect a disabled monitor. 

A good example from my testing is: closed laptop lid + external monitor. In this case, we can't be certain that the probe won't report the internal panel's scale instead of the external monitor.

Once we account for that, this is ready to merge :fire:

ClickSentinel:
@shibco might be good to go now. Was able to capture that case.

shibco:
This is good! Merged!

PR #56, opened 2026-07-23
Add repository Code of Conduct

shibco:
Hi all,
What started as a small rage-driven project early last year is now growing into a multi-contributor open project. I had no idea it would go this way and as the repo grows, so too does the scope and ambition.
Before things get too big, I propose adding a code of conduct that we can use to guide the project. This is forked from the code of conduct qe have used for years at my research organisation, [New Design Congress](https://newdesigncongress.org). It has served us well in the past.
Please share your thoughts or feedback, and lets discuss!

### DIFF
--- a/CODE_OF_CONDUCT.md
+++ b/CODE_OF_CONDUCT.md
@@ -1,0 +1,119 @@
+# Code of Conduct
+
+Para-Real Ltd.
+
+## PREAMBLE
+
+This repository is produced by Para-Real Ltd. This Code of Conduct provides clarity on how we operate it and the standards we hold ourselves and everyone who participates in it to. The project is independent and unofficial; it is not endorsed by or affiliated with Ableton.
+
+The people who participate in this repository are a diverse cohort that asserts that *another future is possible*. They come from a variety of cultural, ethnic and professional backgrounds; the work and the issues we grapple with together are often provocative, emotional and urgent. We are dedicated to building an inclusive and harassment-free environment in this repository for everyone regardless of gender, gender identity and expression, race, ethnicity, sexual orientation, disability, age, religion (or lack thereof), body size, physical appearance, social class or language ability (ability to speak English).
+
+We aspire to live by our values and behave in ways that positively impact our communities, how we treat each other, how we interact with one another, and how we see the world. This Code operationalises our para-real ethos: sovereignty, mutual care, zero extraction. Any behaviour that erodes these principles will be excluded.
+
+This Code covers *everybody* when they participate in this repository, including but not limited to:
+
+- the core Para-Real Ltd. team working on the repository;
+- maintainers, contributors and reviewers;
+- people who open or comment on issues, pull requests or commits;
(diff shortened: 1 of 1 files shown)


yioannides:
I'm okay with the premise; seems fair, well thought out and ironclad. I'll do a second, more thorough pass tonight, but I already feel assured knowing it's a fork from your research org ~

0tanh:
One of the most thorough CoC's I've ever seen. No notes from me <3

CODE_OF_CONDUCT.md
```diff
@@ -0,0 +1,119 @@
...
+
+### Be careful of people's privacy.
+
+Whether working with diagnostic data, test reports, logs, or simply participating in this repository, think before you share. Do not access user data you're not entitled to or disclose information someone provided in confidence. Do not post photos of others online without their consent. Do not share personally identifying information posted by repository participants to other places. Do not post the personal details of others, do not engage in doxing, and do not re-post or re-share links to sites or resources that share or weaponise personal details without first checking with the Para-Real Ltd. team.
+
+Review diagnostic output before publishing it. Do not post Ableton installers, authorisation files, licence keys, projects, samples, recordings, account details or plug-in credentials.
```
jackson-57: I don't see the problem with letting people post projects, samples, or recordings, as long as they own the content!

CODE_OF_CONDUCT.md
```diff
@@ -0,0 +1,119 @@
...
+
+### Communicate thoughtfully.
+
+We encourage people to share their work, interests, related projects, relevant calls and upcoming events when they contribute to the work of the repository. At the same time, this repository is not an advertising timeline / hustlespace. If you wish to promote something, please engage with the repository beyond self interest.
+
+### Respect our work.
+
+Materials in this repository are licensed as stated in the [LICENCE](LICENCE) file and any component-specific notices unless explicitly stated otherwise. We encourage people to share our work, through links, citation and curation, and are open to remix or extension under those licences. Ableton Live itself is proprietary and is not included in this repository. Users must supply their own copy and valid licence. Do not use the repository to share Ableton installers, authorisation bypasses, licence files or credentials.
+
+### Keep our spaces clear of crime and harm.
```
jackson-57, line 63: I feel like the points from here (L63) on down could be made a lot briefer or removed outright ('Refrain from propaganda ops' seems worthy for your group to mention given its wide political scope but is maybe less of a concern for tooling around Ableton on Linux 😃)

CODE_OF_CONDUCT.md
shibco: Ableton specific here meaning IP by Ableton; for example the demo set. But on re-read, I can see how this sounds like "anything made in Ableton" rather than "anything made **BY** Ableton".

I will update the language to confirm this.

CODE_OF_CONDUCT.md
shibco: While in principle I agree, there are examples of bad faith weaponised FOSS, which I want this to cover.

Issue #35, opened 2026-07-20
[UI] Fine-tuning the auto light/dark mode


yioannides:
Thank you so much for all your recent updates! It already looks beautiful, but there are a few things that could improve it:

1. the menu bar would look more unified to the main window if the menu bar inherited the topmost title bar / window color (the `Desktop` parameter), to reference the original design by the Ableton Live team for the Push 3 Linux binary, as seen below (but keeping it as-is is totally fine, of course):

　　<img width="713" height="160" alt="Image" src="https://github.com/user-attachments/assets/f4f39348-91a2-4a7b-9e2c-b6b42a11d888" />

2. reversely, it would be good for the dropdown menu (1) to inherit Live's default lighter dropdown menu color (I think the `Control Text Back` parameter), as seen on the example below (2):

　　<img width="546" height="386" alt="Image" src="https://github.com/user-attachments/assets/38dcb634-e278-4022-92af-c2cda05047b3" />

　　<img width="512" height="488" alt="Image" src="https://github.com/user-attachments/assets/662355d8-2818-4025-8a05-dccff644a844" />

3. I doubt this is possible (I think Wine draws an extra stroke by default?), but _if_ possible, it would be nice for the Wine dropdown menu's unavailable entries (1) to use Live's lighter inactive text color, as seen below:

　　<img width="539" height="660" alt="Image" src="https://github.com/user-attachments/assets/dd5f3f9b-8c26-43bf-8284-40ef8eecb983" />

4. Switching between schemes a. don't happen in real time (within the DAW) b. it takes two reboots of Live to commit, as seen below (this is after closing Live, switching to night mode and re-launching Live):

　　<img width="479" height="234" alt="Image" src="https://github.com/user-attachments/assets/84529e31-04d8-403e-924b-308558adf8d2" />

5. If possible, it would be great if Wine's menu bar & dropdown menus (1) could use the fallback font for non-Latin languages, as seen here with Japanese (2):

　　<img width="550" height="722" alt="Image" src="https://github.com/user-attachments/assets/c95a8e01-4ac3-4ce1-b154-61a44e8fabfd" />

6. extreme nit-pick: is it possible to remove the underlines from the Wine menu bar? It just feels very "I'm running a DAW via Wine" and it would be certainly nice, if there was a hack for it:

　　<img width="512" height="288" alt="Image" src="https://github.com/user-attachments/assets/a2ee7686-3234-4697-97a6-be8ca3bd7b1b" />

---
[...]

PR #72, opened 2026-07-26
Color Theme Sync

ClickSentinel:
## Summary

Fixes issue #35 points 1, 2, 3, 4b, and 6 (menu bar/dropdown theming, disabled-text ghosting, the "stuck on white until two restarts" bug, and the alt-key mnemonic underlines). Point 4a (true real-time switching while Live is open) is partially resolved and documented as a hard ceiling rather than fixed - see below.

- `0049`: drops the double-pass "engraved bevel" draw for grayed menu items. It assumed `COLOR_3DHILIGHT` sits close to the menu background, which holds for the stock light theme but not once the bar is themed dark - both paths now do a single plain draw with `COLOR_GRAYTEXT`, matching what the `MF_HILITE` case already did. Zero diff from stock Wine behavior otherwise, so this was an upstream bug, not fork-specific.
- `0050`: invalidates each process's per-process system-color cache (including already-created brushes/pens) on `WM_SYSCOLORCHANGE`. Wine reads and caches every system color exactly **once** (`get_rgb_entry`'s `loaded` flag) and _never_ re-reads it, so a `SetSysColors` from another process was invisible to an already-running window forever - this is why a scheme change previously needed two full Live restarts to actually show up.
- `0051`: adds `RDW_FRAME` to `NtUserSetSysColors`'s forced repaint so it reaches the non-client area, where the native menu bar actually lives. Lets `setsyscolors.exe` drop its old `EnumWindows` + `DrawMenuBar` workaround entirely - it now just calls `SetSysColors`.
- `0052`: adds `DT_HIDEPREFIX` to the menu bar's `DrawTextW` call (bar only - dropdown items keep their underlines, where Alt-key navigation still applies). Real Windows only shows these once Alt is pressed; this Wine tree never implements that, so the bar always drew them. `DT_HIDEPREFIX` turned out to be a dead flag in this tree already - defined but never checked at the actual underline-draw site in `dlls/user32/text.c`, which only looked at the non-equivalent `DT_NOPREFIX` (shows a literal `&` instead of hiding the underline).

Plus a `theme_watch_loop` (`scripts/ableton-live`) that detects a Preferences change and drives `setsyscolors.exe`, and a `blend_gray_text`/`ableton_newest_prefs_dir` dedup pass shared between the watch loop and the sync path.

Technical implementation: [notes/ABLETON-WINE-MENU-COLOR-THEMING.md](https://github.com/ClickSentinel/ableton-linux/blob/color-theming/notes/ABLETON-WINE-MENU-COLOR-THEMING.md)

## Limitations for point 4

Currently, the Win32 menu bar theme only visually updates when the settings window is **closed**. This happens because the `.cfg` file is only written at that time.

Investigated if Live emits any other signal (file write, registry write, IPC) **while a theme is being live-previewed in the Preferences dialog**, to see if the watch loop could react faster than "on dialog close." Traced with `inotifywait -m -r` across the prefix plus registry-mtime polling during an actual toggle session: **zero external signal of any kind until the dialog closes**. Full writeup in `notes/FINDINGS-LIVE-THEME-PREVIEW-SIGNAL-2026-07-26.md`. Given that, `theme_watch_loop` is already at the practical ceiling for this one - there's nothing to poll or hook that would make it faster.

## Testing

- Full container build (`build.sh`) passes, 72/72 audit checks.
- Manually exercised light/dark toggle plus a mid-session scheme change confirms the menu bar and dropdowns pick up the new colors, greyed items render correctly, and underlines no longer show on the top-level bar (but ensuring dropdown items still show theirs.)
- Confirmed the "stuck on white" no-update-on-restart bug (reported in the issue thread) is also fixed by 0050, independent of the underline/bevel work.

Screenshots: 

image
image


scripts/ableton-live
```diff
@@ -407,15 +421,28 @@ detect_system_topbar() {   # sets topbar from the host titlebar colors
...
+    local applied=""    # last color set actually broadcast; skip re-firing when unchanged
+    local major have_inotify=0 f
+    local -a watch_files
     [ -f "$apply" ] || return 0
     command -v flock >/dev/null 2>&1 || return 0
+    # inotifywait (inotify-tools) turns this into an event-driven wait instead
+    # of a fixed poll: wakes up on the write itself rather than up to 2s later.
+    # Optional - falls back to the plain sleep-poll below when not installed,
+    # same as the flock check above already does for its own dependency.
+    command -v inotifywait >/dev/null 2>&1 && have_inotify=1
```
shibco, line 444: if this fails (eg on L461 and below), this will ignore the inotifywait exit status entirely (>/dev/null 2>&1, no fallback). On an actual error it returns exit 1 immediately, and the loop then iterates with no sleep at all. Not the worst, but accounting for the loop failure is ideal.

shibco (approved):
One small change to account for inotifywait failures, and this is good to merge.

scripts/ableton-live
ClickSentinel: Good catch! Will fix and merge. Thank you!

Issue #49, opened 2026-07-21
Crackling audio with PipeASIO


yioannides:
https://github.com/user-attachments/assets/2f496aef-4f3b-480e-b398-521076b0edf4

I have been getting crackling audio with these few last updates, I am not sure why... Increasing / decreasing the buffer size doesn't seem to change much...

If there's anything you'd like me to try, please feel free to let me know!

-----

OS: Fedora Linux 44 (Workstation Edition) / Wayland
Kernel: 7.1.4-200.fc44.x86_64
DE: GNOME Shell 50.3
CPU: AMD Ryzen 7 7700X 8-Core Processor
GPU1: Advanced Micro Devices, Inc. [AMD/ATI] Navi 23 [Radeon RX 6600/6600 XT/6600M] (rev c7)
GPU2: Advanced Micro Devices, Inc. [AMD/ATI] Raphael (rev c3)

---
[Discord thread](https://discord.com/channels/1307779762189504602/1530636099393945611/1530636099393945611)

shibco:
Wow that sounds like ass 👀

yioannides:
Facts

yioannides:
This is the output of `pipewire.service`, `wireplumber.service` and `pipewire-pulse.service` after a fresh reboot:

```
Jul 22 08:53:39 fedora systemd[1571]: Started pipewire.service - PipeWire Multimedia Service.
Jul 22 08:53:39 fedora systemd[1571]: Started wireplumber.service - Multimedia Service Session Manager.
Jul 22 08:53:40 fedora systemd[1571]: Started pipewire-pulse.service - PipeWire PulseAudio.
Jul 22 08:53:41 fedora wireplumber[1871]: [0:00:14.241248450] [1871]  INFO Camera camera_manager.cpp:340 libcamera v0.7.1
Jul 22 08:53:41 fedora wireplumber[1871]: [0:00:14.265826951] [3034]  WARN V4L2 v4l2_pixelformat.cpp:346 Unsupported V4L2 pixel format H264
Jul 22 08:53:41 fedora wireplumber[1871]: [0:00:14.265889421] [3034]  INFO Camera camera_manager.cpp:223 Adding camera '\_SB_.PCI0.GP17.XHC0.RHUB.PRT2-2:1.0-046d:082d' for pipeline handler uvcvideo
[...]

PR #73, opened 2026-07-26
Document the PipeASIO crackle causes and fix plan

shibco:
I spent the last couple of days researching this and making notes on #49. I believe these are two separate issues that despite having two differing root causes nevertheless have the same outcome.

Of note, I've done some deeper research into the decisions Ableton's engineering team have made around how the audio engine responds to various host contexts, and this has been illuminating. More than anything, I don't think it can be put down to a single specific fix yet, rather the crackling is the result of multiple issues converging, which is also possibly why its so hard to reproduce on either my main machine or the various testing environments we have.

@ClickSentinel, I'm documenting this here as a draft PR, and plan to take a look at it on and off tomorrow during downtime at my day job. I want to fix this as a priority, as obviously "shitty audio" is the absolute worst outcome for anyone looking to switch. Please feel free to take a stab at some of this yourself if you find the time and drop any findings you have here. 

Right now, this contains a single commit (the documentation of my research so far), but will expand into an actual patch and PR when we produce a fix.

### DIFF
--- a/notes/ABLETON-WINE-PIPEASIO-CRACKLE.md
+++ b/notes/ABLETON-WINE-PIPEASIO-CRACKLE.md
@@ -1,0 +1,320 @@
+# PipeASIO crackle: causes and fix plan
+
+Status: plan, 2026-07-26. Nothing implemented.
+
+This note identifies the faults behind the crackle reports, the fix for each,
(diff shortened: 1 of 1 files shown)


shibco:
This should be fixed as a result of the migration to newer pipeasio

Issue #61, opened 2026-07-25
Live randomly freezes with the desktop shell


NoskyD:
## What happened

I'm on CachyOS with KDE Plasma using 12 Suite. 
Ableton Live tends to freeze the desktop environment (including the mouse pointer) in a random moment to the state when only hard reset or Alt + SysReq + REISUB helps. The audio engine proceeds to work and accept notes from the MIDI keyboard. (The expression control effect stops working too)

I've ran the realtime setup script, but it didn't help much.

## Steps to reproduce

1. Run Ableton Live from the application launcher.
2. Make a mildly intensive situation (i.e. change knobs on effects while playing the song).
3. Wait until Live and your desktop environment freezes.

## System summary

Run the profiler for the machine you saw the problem on. It prints a redacted system summary and copies it to your clipboard.

You can also share additional information - like what your working setup looks like - by running the macOS or Windows profiler and adding this to the issue as well.

- Linux: `./beta/scripts/ableton-linux-profiler.sh`
- macOS: `./beta/scripts/ableton-macos-profiler.sh`
- Windows: `powershell -ExecutionPolicy Bypass -File beta\scripts\ableton-windows-profiler.ps1`

Paste the summary here:

*Didn't find that script. I only have an Ableton's crash report.*

---
[Discord thread](https://discord.com/channels/1307779762189504602/1530636225336316125/1530636225336316125)

NoskyD:
I've tried out the new release. Live doesn't seem to freeze anymore! 
Thanks for fixing it.

shibco:
<kbd>[![NoskyD 7\](https://cdn.discordapp.com/avatars/948406592552988753/513527f09cf8ae6ace74b9d4a054331f.webp?size=40)](https://discord.com/channels/1307779762189504602/1530636225336316125/1532233764158312509)</kbd> [NoskyD 7\](https://discord.com/channels/1307779762189504602/1530636225336316125/1532233764158312509)  `BOT`


![imogen-heap-big-hug.gif](https://cdn.discordapp.com/attachments/1530636225336316125/1532233763847798985/imogen-heap-big-hug.gif?ex=6a6c1b64&is=6a6ac9e4&hm=855319acea0527a3f516b4e337e3bb325a2d539aea1e6282641f4a11408e6569& "imogen-heap-big-hug.gif")

shibco:
[...]

Issue #57, opened 2026-07-23
Learn View close leaves a permanent 5 Hz flicker in the old pane rect (0041's teardown-only reblit liveness vs Live's parked WebView2 stack)


jttdev:
### Symptom

On 2026.07.21.2 through 2026.07.23.1 (i.e. with patch 0041), **closing** the Learn View leaves a steady flicker in the screen region the pane occupied: the last Learn View frame alternates with Live's own UI, indefinitely, until Live is restarted. Reproduced on Live 12.4.5b7, 125% scale, KDE/X11. The pane itself (open, heal, hold) works exactly as 0041's notes describe — this is a close-path gap, not a regression of the heal.

### Root cause

Live does not tear the WebView2 stack down on pane close — it **parks it for reuse**. `hwndspy` during the flicker:

```
000000000001009C  "Ableton Live Window Class"  "Unbenannt - Ableton Live 12 Beta"  VIS
00000000000300E2  "AbletonWebViewHelperWindow"                                     hidden
  00000000000100E6  "Chrome_WidgetWin_0"                                           hidden
    0000000000010152  "Chrome_WidgetWin_1"  "Learn View 12"                        hidden
      00000000000101A4  "Intermediate D3D Window"  rect 907x1523  VIS  ex=00280024 LAYERED
```

The pane chain is reparented under a hidden `AbletonWebViewHelperWindow`; the browser and GPU processes stay alive, the composition swapchain is **not** Released, and the dcomp target window survives — still `WS_VISIBLE` at the old pane size, every ancestor hidden.

Patch 0041's reblit liveness is teardown-only (`d3d11_swapchain_Release` / `WM_NCDESTROY`), and it removed 0025's `DCOMP_STALE_TICKS` idle suspension. In the parked state neither teardown signal ever fires, so the 200 ms idle reblit keeps stamping the last comp buffer into the host toplevel's surface at the stale pane offset forever:

- `+dxgi` trace: `Re-blit #400: hwnd 00000000000101A4 907x1523 reason=timer-idle` still ticking minutes after the close (GPU process). `WM_PAINT`s stop at close (hidden windows receive none) — only the timer painter survives.
- XDamage on Live's toplevel, 3 s window: 15× damage of `906x1523` at the old pane offset (= 5 Hz, the reblit cadence) interleaved with Live's own ~50 Hz full-client presents. That alternation is the visible flicker.

[...]

Issue #42, opened 2026-07-21
Full screen option broken


jackson-57:
View->Full Screen or F11 should put Live in full screen and hide the title bar, menu bar, and task bar.
Currently, if you use View->Full Screen, only the title bar is hidden. If you use F11, the title bar and menu bar are hidden, but the window is corrupted and shifted.

https://github.com/user-attachments/assets/8fea9e8a-e2d5-4de8-9c24-2b9b98cba091

---
[Discord thread](https://discord.com/channels/1307779762189504602/1530635973375824115/1530635973375824115)

shibco:
fixed in the forthcoming `2026.07.21.1` release <3

jackson-57:
we chatted about this on discord, this one unfortunately is not fixed with KDE :(

shibco:
🥲

ClickSentinel:
Also able to replicate this against main :( #74 does not fix unfortunately.

PR #74, opened 2026-07-27
Fix/interface drawing on gpu

shibco:
So after a maddening few days struggling with WebView2 artifacting and flickering, I'm pleased to share that the fix was very simple.

Early in the development of this project, I wrote off GPU support entirely for simplicity's sake. I was solving for larger issues with Live that (until now! :fire:) were almost universal experiences when trying to get Live 12 running via Wine: the UI click targets not lining up with the UI itself, Live crashing or suffering violent, seizure-inducing full screen flickering on the main window and/or modals, inability to resize at all, massive misalignment on DEs with fractional scaling, etc.

But not all of these were caused by the GPU, and regardless of whether the larger issues were caused by the GPU or not, all of these problems were fixed before I published the project. It now turns out that we are good to go with GPU support, and as such the fix for the flicker and much (not all) of the UI instability around the Splice and Learn views is resolved entirely by removing `-_ForceGdiBackend` from `Options.txt`, allowing Live to switch to its Direct2D/D3D11 renderer.

On Windows, apps like Live and Chromium (which draws the WebView2 panes) don't put pixels on screen themselves. They hand finished frames to the Windows compositor, DirectComposition, and Windows puts them on screen. Wine has no real version of that compositor. This Wine fork fills the gap on a timer that copies the latest WebView2 frame into the pane window so the content stays visible.

But this hack is a hack that also sucks: Chromium _also_ paints its own software frame into the same window. With Live on the old GDI renderer, the two frames held different pictures, one current and one stale, so the pane switched between them about 5 times a second. That switching is the flicker, and the 5 Hz rate is exactly the copy timer. With Live on its GPU renderer, both painters end up drawing the same picture rather than mangling it. The switching still happens, but you can no longer see it.

One additional side effect that's so awesome to share: Measured with `XDamage` capture and frame hashing on both the 11.11 production runtime and the 11.13 build of main, at 200% scale, on WebView2 `149,` and confirmed in use: **Idle CPU also drops to 1-2%.** So good.

We also update the Prefix setup (step 5c) to remove the old Options.txt line from existing prefixes.

Turning the GPU renderer on exposed one more bug, which this PR also fixes. When you drag-resize a window on Windows, the system asks the app for its minimum size and stops the drag there. Wine never told the Linux window manager what Live's minimum size is, so the window manager happily made the window smaller than Live allows. Live pushed back mid-drag with its minimum size, the window manager applied that push at the edge you were holding, and the window grew away from your drag, sometimes ending with the window suddenly maximizing itself. Patch `0053` makes Wine pass the app's minimum size to the window manager as the standard X11 `PMinSize` hint, so the drag now just stops at Live's minimum, the same behaviour as on Windows.

Also adds `liveinject ctrlalt7` (scancode injection), used during diagnosis. 
You can find my full, frantic notes at `notes/WEBVIEW2-DIAGNOSIS.md` and the breakthrough at `notes/ABLETON-WINE-GPU-RENDERER.md`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,15 @@
 
 ## Unreleased
 
+- Live now uses its GPU renderer. Prefix setup removes the legacy
+  `-_ForceGdiBackend` line from `Options.txt` (step 5c). This removes the
+  Learn View and Splice view flicker and drops idle CPU to 1-2%. See
+  [the GPU renderer note](notes/ABLETON-WINE-GPU-RENDERER.md).
+- Fixed windows fighting an interactive resize below the app minimum
+  (Wine patch 0053). winex11 now exports the `WM_GETMINMAXINFO` minimum
+  as the X11 `PMinSize` hint, and the window manager stops the drag at
+  the minimum.
+
 - Fixed a Live crash when closing WebView2 plugin editors (issue 52, Wine
   patch 0045). `RevokeDragDrop` now rejects windows owned by another process,
   matching `RegisterDragDrop`. Fix by Giang Nguyen. See
--- a/WEBVIEW2-DIAGNOSIS.md
+++ b/WEBVIEW2-DIAGNOSIS.md
@@ -1,0 +1,213 @@
+# WebView2 Learn View flicker: diagnosis and fix plan
+
+2026-07-27. Covers the open-pane Learn View artifact reported on the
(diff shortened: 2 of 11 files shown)


shibco:
Additionally, this could possibly partially solve or completely resolve #61, #57, #49 and #42

ClickSentinel (approved):

PR #76, opened 2026-07-27
Add Wine powershell.exe shim helper for installers that use PowerShell preflight loops

midisaur:
see https://github.com/midisaur/ableton-linux/pull/1 — opened this on the fork; trying to surface it to the upstream for review.

### DIFF
--- a/scripts/setup-powershell-shim.sh
+++ b/scripts/setup-powershell-shim.sh
@@ -119,15 +119,26 @@ patch_one "$WINE_ROOT/lib/wine/x86_64-windows/powershell.exe" "wine-loader-64"
 patch_one "$WINEPREFIX/drive_c/windows/system32/WindowsPowerShell/v1.0/powershell.exe" "prefix-system32"
 patch_one "$WINEPREFIX/drive_c/windows/syswow64/WindowsPowerShell/v1.0/powershell.exe" "prefix-syswow64"
 
-# 3) Sanity check the loader stub. This is the one the installer actually
-#    hits via Wine's builtin dispatcher, so we verify it directly.
+# 3) Functional sanity check. The shim's matching logic returns 1 only when
+#    argv contains all three substrings: Get-CimInstance, Win32_Process,
+#    StartsWith. We feed it a script that contains those substrings (mimics
+#    the Native Access 2 preflight), so the shim should match and exit 1.
+#    If it exits 0 instead, the matching code path didn't fire — the patch
+#    did not apply, or a stale shim from a previous Wine tree is still in
+#    place. This catches both failure modes; an unrelated script (like
+#    'exit 7') would only test "the shim runs at all", not "the matching
+#    logic works", which is what we actually want to verify.
 echo
-echo "Sanity: $WINE_ROOT/bin/wine powershell.exe -C 'exit 7'"
-if WINEPREFIX="$WINEPREFIX" "$WINE_ROOT/bin/wine" powershell.exe -C 'exit 7' >/dev/null 2>&1; then
-    echo "   WARNING: shim returned 0 for an unrelated script. Patch may not have applied."
+echo "Sanity: $WINE_ROOT/bin/wine powershell.exe -C '<Get-CimInstance ... StartsWith script>'"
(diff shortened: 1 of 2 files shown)


scripts/setup-powershell-shim.sh
```diff
@@ -0,0 +1,155 @@
...
+
+# 3) Sanity check the loader stub. This is the one the installer actually
+#    hits via Wine's builtin dispatcher, so we verify it directly.
+echo
+echo "Sanity: $WINE_ROOT/bin/wine powershell.exe -C 'exit 7'"
+if WINEPREFIX="$WINEPREFIX" "$WINE_ROOT/bin/wine" powershell.exe -C 'exit 7' >/dev/null 2>&1; then
+    echo "   WARNING: shim returned 0 for an unrelated script. Patch may not have applied."
+    exit 2
+fi
+echo "   OK: shim returned 1 (mimics the original stub's 'do nothing, fail' for any -C)."
```
ClickSentinel: Might need to review this sanity check. Exit 7 matches none of the 3 substrings. Correctly working shim should return 0 as per L86-L87

```c
if (found_cim && found_proc && found_starts) return 1;
return 0;  /* everything else: behave like the original Wine stub */
```

In bash, `if COMMAND; then` fires on exit code 0, so a correctly-installed shim may trigger the `WARNING`/`exit 2` branch every time if i'm reading that right? 

The "OK: shim returned 1" message can only ever print if Wine itself fails to even launch powershell.exe (a crash or missing-file error) - not because the shim's matching logic ran correctly.

scripts/setup-powershell-shim.sh
```diff
@@ -0,0 +1,155 @@
...
+    }
+    /* Standard NA-style process-already-running check.
+     * We deliberately don't gate on the app name (e.g. "Native Access")
+     * so the same shim works for any installer that uses the same
+     * Get-CimInstance + Win32_Process + StartsWith pattern. */
+    if (found_cim && found_proc && found_starts) return 1;
+    return 0;  /* everything else: behave like the original Wine stub */
+}
+EOF
+x86_64-w64-mingw32-gcc -O2 -s -o "$shim_bin" "$shim_src"
```
ClickSentinel, line 90: We could probably use the no-CRT approach instead of introducing a new toolchain. Other compiled helpers use `clang` building a no-CRT PE binary via `build_*.sh` scripting in the main build.

shibco (changes requested):
@midisaur thanks for this, and for your investigative work 

This deals with a real problem. Wine's `powershell.exe` is a stub that ignores `-C` and always returns `0`, so any installer that uses a PowerShell preflight to check whether it is already running will loop forever. We need to fix this.

Some feedback before merge:

- As @ClickSentinel raised, the sanity check is inverted. Right now, the shim returns `0` for anything that does not match all three substrings. The check runs `powershell.exe -C 'exit 7'`, which does not match, so the shim returns `0` and the script exits `2` with a warning after a correct patch. As written it can never reach the "Done" message. It also means a shim that failed to build gets reported as working.

- The shim is built once with `x86_64-w64-mingw32-gcc` and copied over all three paths. But the `syswow64` copy is 32-bit. This PR would need to script a second build with `i686-w64-mingw32-gcc`.

- As this stands, we won't be able to merge this, because this requires the user have the compiler on their machine, which drastically increases the risk of install failure. The approach we've taken here is to produce a compiled Wine fork with patches layered sequentially via `scripts/install.sh`.  The idea here is to rework this as `patches/0053-...` along with the matching entry in `scripts/build-audit.sh`, following the convention of the project, rather than a live compilation task.

We're going to get a lot of requests for NI support, so I'm very grateful you submitted this!

ClickSentinel (changes requested):
A few things I would like to see before we can pull this in:

- Doesn't currently build: mktemp gives a suffix-less path, and GCC needs a .c extension (or -x c) to treat it as C source. As written, the build step fails with ld: file format not recognized every time.
- Patching in place is fragile: overwriting files under $WINE_ROOT/lib/wine/... and the prefix's system32/syswow64 means a Wine update or a fresh setup-prefix.sh run silently reverts the patch with no warning. Could this use Wine's own override mechanism (WINEDLLOVERRIDES / HKCU\Software\Wine\DllOverrides) instead of mutating files Wine manages itself?
- Nitpicking: scripts/setup-powershell-shim.sh is missing its executable bit in this diff.

PR #77, opened 2026-07-27
Patch 0054: Language fallback for non-latin characters

### LINKED ISSUE #35: [UI] Fine-tuning the auto light/dark mode

ClickSentinel:
## Summary

Fixes #35 point 5: non-Latin menu text — a non-Latin locale, a translated menu, a non-Latin project or track name — rendered as tofu or nothing at all, because the launcher substitutes Wine's stock chrome font with Ableton's own `Ableton Sans`/`Ableton Sans Small`, which is Latin-only.

## The fix

Wine populates the `SystemLink` fallback chain correctly but consults it only when selecting a whole font by charset, never when rendering a glyph, so a registered fallback family sits unused. `0054` adds the missing per-string consumer in `dlls/win32u/menu.c`: `draw_menu_item` tests coverage with `NtGdiGetGlyphIndicesW` and, on a miss, walks the registered families (`get_font_link_families`, new in `font.c`) for the first that covers the string.

`calc_menu_item_size` now measures with that same font. It had sized `item->rect` with the original, missing-glyph font, so the wider fallback render was clipped — `"オプション"` truncating to `"プショ"`.

Three smaller bugs found while tracing real Japanese text. `NtGdiExtGetObjectW` returns the resolved face's full name rather than the family the chain is registered under, so the style suffix is stripped and the lookup retried. Item text carries its `'\t'`-separated shortcut column and no font has a glyph for a tab, which was failing otherwise-covered entries. Lone UTF-16 surrogate halves are in no font's cmap, so a Set name holding an emoji failed every candidate on every paint.

The candidate array was `WCHAR names[4]`, so only the first four chain entries were ever tried. The chain the launcher builds is six, leaving both Unifont entries — the safety net the code documents — unreachable. Raised to 16.

## Fallback fonts

`sync_font_fallback` originally picked the CJK fallback with `fc-match` against whatever the host had installed, so the result varied per distro. It now reads known files out of Live's `Resources/Fonts` directly with `fc-scan`. Live ships Noto Sans CJK because its own UI is localized into Japanese and Chinese, so the files are there whenever Live is, and `sync_ui_font`'s auto path already returns early unless that directory exists. Both weights are linked, so bold menu items get a real bold face. Nothing is vendored and the release artifact does not grow.

Entries use `%{fullname}`, not `%{family}`. `load_system_links` matches a `SystemLink` entry on filename plus the **name-ID-1** family, and `%{family}` reports the ID 16 typographic name first — Live's `NotoSansCJKjp-Regular.otf` is `"Noto Sans CJK JP Regular"` in ID 1, with the bare `"Noto Sans CJK JP"` only in ID 16. The ID 16 name matched no family win32u knew, so the entry was dropped at load with a silent `Unable to find file` and Japanese fell through Tahoma to Unifont: rendering, but blocky.

## Testing

- [x] Full container build passes, 76/76 audit checks.
- [x] Japanese menu bar and dropdowns render correctly end to end, including the previously-truncated settings label. @amenohi2 tested an earlier revision against a native macOS build and supplied the screenshots that confirmed the layout — thanks.
- [x] Four-entry cap verified rather than assumed: with host CJK hidden via `FONTCONFIG_FILE`, no Noto in the prefix, and a chain crafted so the only covering font sat at position five, Japanese rendered — unreachable under the old cap.
- [x] Name-ID-1 fix proven causally in a prefix built with no host fonts at all, one variable changed: the ID 16 name gives `Unable to find file`, the ID 1 name loads Live's own file. The fresh prefix matters — `FONTCONFIG_FILE` alone cannot isolate one that has already persisted the host's font paths into its registry, which quietly invalidated two earlier comparisons.

`notes/ABLETON-WINE-MENU-FONT-FALLBACK.md` documents the mechanism and is tracked here for the first time.

The commits @shibco reviewed are no longer on any branch, though GitHub still resolves them by hash, so the `Fixed in <sha>` links in my replies above still work.

### DIFF
--- a/patches/0054-win32u-fall-back-to-a-linked-font-for-missing-glyph.patch
+++ b/patches/0054-win32u-fall-back-to-a-linked-font-for-missing-glyph.patch
@@ -1,0 +1,323 @@
+Subject: win32u: fall back to a linked font when the menu font is
+ missing a glyph
+
+The launcher substitutes Wine's stock win32 chrome font (menu bar, dropdowns,
+dialogs) with Ableton's own UI face (Ableton Sans/Ableton Sans Small), which
+covers Latin script only - checked directly via its cmap tables, zero glyphs
+anywhere in Cyrillic, Arabic, Hebrew, Devanagari, Thai, CJK, or Hangul. Any
+menu text needing a glyph outside that (a non-Latin locale, a Japanese
+project/track name, ...) was simply missing entirely: tofu boxes, or nothing
+at all.
+
+Wine's own font-linking (HKLM\...\FontLink\SystemLink, correctly populated
+by load_system_links - confirmed by trace) turns out to only ever be
+consulted during whole-font *selection* by system charset
+(can_select_face), never during glyph *rendering* - so registering a
+fallback family there fixes nothing on its own; the raw glyph lookup
+(freetype_get_glyph_index) only ever checks the one font actually selected,
+with no awareness of any linked fallback at all. This is the missing
+per-string consumer: if the currently selected font can't draw the item's
+text, walk its registered SystemLink fallback families (get_font_link_families,
+font.c - new, since find_gdi_font_link and the data it tracks were
+otherwise only reachable from within font.c) and temporarily select the
+first candidate that actually covers the text, sized and weighted to
+match, for the DrawTextW calls that follow.
+
+Measuring, not just drawing, needs to agree on which font is in play too:
+calc_menu_item_size sizes item->rect by measuring item->text with whatever
+font is already selected in hdc - the original (missing-glyph) font, not
+the fallback font just described. A fallback font's real glyphs are
+typically wider than the original font's notdef/tofu metrics, so
+item->rect ended up sized for the wrong font, and the DrawTextW call above
+then clipped the wider real render to fit - CJK menu items losing their
+first and last characters (menu bar, centered) or losing text off the
+right edge (popup items, left-aligned). Fixed by forward-declaring
+get_fallback_font_for_text and calling it in calc_menu_item_size the same
+way, before measuring instead of before drawing; both call sites already
+select the same plain (non-bold) menu font into hdc before their
+respective loops, so the two selections stay in lockstep.
+
+Whole-string swap, not per-character: text mixing scripts renders entirely
+in the fallback font rather than mixing faces mid-string. Real per-glyph
+fallback is a Uniscribe/DirectWrite-level feature that plain DrawTextW
+(what this file actually calls) has never had; implementing that properly
+is a much larger undertaking than this - this trades a rarer, minor
+font-consistency slip for the text actually rendering at all.
+
+Three bugs found and fixed along the way:
+- NtGdiExtGetObjectW returns the *resolved face's* full name (e.g.
+  "Ableton Sans Small Regular"), not the family name the SystemLink chain
+  is registered under ("Ableton Sans Small") - strip the known style
+  suffix and retry. Confirmed by tracing real Japanese menu text.
+- The item text passed in includes its raw '\t'-separated shortcut column
+  (see draw_menu_item); no font has an actual renderable glyph for a
+  literal tab, so it was being flagged as "missing" in every candidate
+  font tried, failing entries that were otherwise fully covered. Control
+  characters are no longer treated as glyphs needing coverage. Confirmed
+  by tracing real Japanese menu text.
+- font_missing_glyphs works per UTF-16 code unit, and a lone surrogate
+  half (0xD800-0xDFFF) is never in any font's cmap - not because the
+  glyph is missing, but because a surrogate half isn't a real character
+  on its own. Without a skip for it, any menu item holding a character
+  outside the BMP (an emoji in a Set name, say) reported "missing"
+  against the item's own font and every fallback candidate, always - a
+  guaranteed-to-fail check re-run on every single paint, in both
+  calc_menu_item_size and draw_menu_item. Surrogates get the same skip
+  control characters already get; whole-string fallback can't fix a lone
+  surrogate regardless of which font is picked, so there is nothing to
+  gain by still running the check for one.
+
+Verified live end to end: Japanese menu bar and dropdown text, including a
+previously-truncated settings label, renders in full.
+---
+ dlls/win32u/font.c           |  27 ++++++
+ dlls/win32u/menu.c           | 179 +++++++++++++++++++++++++++++++++++++++-
+ dlls/win32u/win32u_private.h |   1 +
+ 3 files changed, 205 insertions(+), 2 deletions(-)
+diff --git a/dlls/win32u/font.c b/dlls/win32u/font.c
+index dcc75e8..f1bebe6 100644
+--- a/dlls/win32u/font.c
++++ b/dlls/win32u/font.c
+@@ -1460,6 +1460,30 @@ static struct gdi_font_link *find_gdi_font_link( const WCHAR *name )
+     return NULL;
+ }
+ 
++/* Copies up to max SystemLink fallback family names registered for name into
++ * out, returns how many were copied. Plain accessor exposing what
++ * find_gdi_font_link tracks; callers do their own coverage testing and font
++ * creation through the public syscalls. */
++UINT get_font_link_families( const WCHAR *name, WCHAR out[][LF_FACESIZE], UINT max )
++{
++    struct gdi_font_link *font_link;
++    struct gdi_font_link_entry *entry;
++    UINT n = 0;
++
++    pthread_mutex_lock( &font_lock );
++    if ((font_link = find_gdi_font_link( name )))
++    {
++        LIST_FOR_EACH_ENTRY( entry, &font_link->links, struct gdi_font_link_entry, entry )
++        {
++            if (n >= max) break;
++            lstrcpynW( out[n], entry->family_name, LF_FACESIZE );
++            n++;
++        }
++    }
++    pthread_mutex_unlock( &font_lock );
++    return n;
++}
++
+ static struct gdi_font_family *find_family_from_font_links( const WCHAR *name, const WCHAR *subst,
+                                                             FONTSIGNATURE fs )
+ {
+diff --git a/dlls/win32u/menu.c b/dlls/win32u/menu.c
+index 1ff3403..a054d46 100644
+--- a/dlls/win32u/menu.c
++++ b/dlls/win32u/menu.c
+@@ -2006,6 +2006,11 @@ static void get_bitmap_item_size( struct menu_item *item, SIZE *size, HWND owner
+     }
+ }
+ 
++/* Defined below, needed here too: calc_menu_item_size must measure text with
++ * the same font draw_menu_item will actually render it in (see the call site
++ * further down for why). */
++static HFONT get_fallback_font_for_text( HDC hdc, const WCHAR *text, INT count );
++
+ /* Calculate the size of the menu item and store it in item->rect */
+ static void calc_menu_item_size( HDC hdc, struct menu_item *item, HWND owner, INT org_x, INT org_y,
+                                  BOOL menu_bar, struct menu *menu )
+@@ -2104,12 +2109,18 @@ static void calc_menu_item_size( HDC hdc, struct menu_item *item, HWND owner, IN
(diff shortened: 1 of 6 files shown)


scripts/ableton-live
```diff
@@ -588,6 +588,47 @@ sync_font_substitutes() {   # family -> route the shell-dialog faces to it
...
+# *selection* by system charset, never during actual glyph *rendering* - see
+# get_fallback_font_for_text in dlls/win32u/menu.c for the real consumer;
+# this is just the data source it reads.
+sync_font_fallback() {   # family -> link Tahoma (+ a host CJK font, if any) as its SystemLink fallback
+    local fam="$1" want="tahoma.ttf,Tahoma" cur
+    local winfonts="$WINEPREFIX/drive_c/windows/Fonts" cjk_file cjk_fam
+    if command -v fc-match >/dev/null 2>&1; then
+        cjk_file="$(fc-match -f '%{file}' ':lang=ja' 2>/dev/null)"
+        cjk_fam="$(fc-match -f '%{family}' ':lang=ja' 2>/dev/null | cut -d, -f1)"
+        if [ -n "$cjk_file" ] && [ -n "$cjk_fam" ] && [ -r "$cjk_file" ]; then
```
shibco: `fc-match` always returns a font. It returns a best match even when nothing on the system satisfies the pattern, so this guard never fires. On a `fontconfig` root holding only DejaVu Sans:

`fc-list  :lang=ja` -> 0 results
`fc-match :lang=ja` -> .../DejaVuSans.ttf|DejaVu Sans

On any computer without a CJK font, and that is a lot of default installs, every launch symlinks an unrelated host font into `drive_c/windows/Fonts` and registers it as a `SystemLink` fallback. `fc-list :lang=ja` returns nothing when the host genuinely has no Japanese font, so we need that check there.

scripts/ableton-live
```diff
@@ -588,6 +588,47 @@ sync_font_substitutes() {   # family -> route the shell-dialog faces to it
...
+# Ableton Sans/Ableton Sans Small cover Latin script only (checked directly:
+# 0 glyphs anywhere in Cyrillic, Arabic, Hebrew, Devanagari, Thai, CJK, Hangul
+# - confirmed via their cmap tables). Register a SystemLink fallback chain for
+# the substituted family so win32u/menu.c's get_fallback_font_for_text
+# (patch 0054) has something to walk when a menu string needs a glyph the
+# primary font doesn't have - Tahoma (Wine's only bundled font with real
+# non-Latin coverage: checked, 104 Cyrillic + 162 Arabic glyphs), plus
+# whatever real CJK font the host happens to have installed for its own
+# desktop, since Wine bundles none. Best-effort: silently skipped if
+# fc-match is missing or the host genuinely has no CJK font at all - same
```
shibco: This is not what happens. See the note on L613 and update the comment needs to match the behaviour once the check is fixed.

patches/0054-win32u-fall-back-to-a-linked-font-for-missing-glyph.patch
```diff
@@ -0,0 +1,332 @@
...
++    BOOL missing = FALSE;
++    INT i;
++
++    if (count > (INT)ARRAY_SIZE(stack_buf) && !(buf = malloc( count * sizeof(*buf) )))
++        return FALSE;   /* can't check: leave the font alone rather than guess */
++
++    if (NtGdiGetGlyphIndicesW( hdc, text, count, buf, GGI_MARK_NONEXISTING_GLYPHS ) != GDI_ERROR)
++    {
++        for (i = 0; i < count; i++)
++            if (buf[i] == 0xffff && text[i] >= ' ') { missing = TRUE; break; }
```
shibco: `NtGdiGetGlyphIndicesW` works per UTF-16 code unit, and a lone surrogate half is in no font's `cmap`. Surrogates will pass this test, so any menu item holding a character outside the BMP reports missing in the item font and in all four candidates. That runs four `NtGdiHfontCreate` plus select, test, and delete cycles per item, in both `calc_menu_item_size` and `draw_menu_item`, on every paint, and it can never succeed. An emoji in a Set name under File > Recent Sets is enough to trigger it.

Surrogates want the same skip the control characters get, and whole-string fallback cannot help them either way. This one's a blocker for sure.

scripts/ableton-live
```diff
@@ -588,6 +588,47 @@ sync_font_substitutes() {   # family -> route the shell-dialog faces to it
...
+# get_fallback_font_for_text in dlls/win32u/menu.c for the real consumer;
+# this is just the data source it reads.
+sync_font_fallback() {   # family -> link Tahoma (+ a host CJK font, if any) as its SystemLink fallback
+    local fam="$1" want="tahoma.ttf,Tahoma" cur
+    local winfonts="$WINEPREFIX/drive_c/windows/Fonts" cjk_file cjk_fam
+    if command -v fc-match >/dev/null 2>&1; then
+        cjk_file="$(fc-match -f '%{file}' ':lang=ja' 2>/dev/null)"
+        cjk_fam="$(fc-match -f '%{family}' ':lang=ja' 2>/dev/null | cut -d, -f1)"
+        if [ -n "$cjk_file" ] && [ -n "$cjk_fam" ] && [ -r "$cjk_file" ]; then
+            ln -sf "$cjk_file" "$winfonts/${cjk_file##*/}" 2>/dev/null
```
shibco: This runs before the `[ "$cur" = "$want" ]` early return on line 626, so it writes on every launch even when nothing changed. Looks like it will also overwrite a real file of the same name in the prefix Fonts directory.

There is no prune that accounts for this link later either. `sync_ui_font` prunes stale AbletonSans* links, but the same behaviour is not implemented for this one.

scripts/ableton-live
```diff
@@ -626,6 +667,7 @@ sync_ui_font() {
     fi
     sync_font_substitutes "$family"
     sync_metric_fonts "$family"
+    sync_font_fallback "$family"
```
shibco, line 699: The `ABLETON_UI_FONT=off` path will return before this, so it never clears the SystemLink value it wrote earlier. Turning the substitution off leaves the fallback chain registered for a family that is no longer in use.

patches/0054-win32u-fall-back-to-a-linked-font-for-missing-glyph.patch
```diff
@@ -0,0 +1,332 @@
...
+         if (item->fState & MFS_DEFAULT)
+              prev_font = NtGdiSelectFont( hdc, get_menu_font(TRUE) );
+ 
++        /* A fallback-rendered run (see get_fallback_font_for_text below) is
++         * typically wider than the original font's missing-glyph metrics -
++         * measuring here with the original font sizes item->rect too small,
++         * and draw_menu_item's later render in the fallback font then gets
++         * clipped by DrawTextW to that too-small rect (reported as CJK menu
++         * text losing its first/last characters, issue #35). */
++        if ((fallback_font = get_fallback_font_for_text( hdc, item->text, -1 )))
```
shibco, line 142: Noto Sans CJK is about 21% taller than Ableton Sans Small at the same size (win ascent plus descent, 1.448 em against 1.200 em). At the prefix `MenuFont` height of `-11` that moves `txt_height` from about 13px to about 16px. Patch `0029` codes the menu bar band at `SM_CYMENU + 3`, which is roughly 21px, so it still fits, but only by about 5px.

This is _probably_ ok, but we should also _probably_ do an interactive resize with Japanese menu text at 200% scale. This menu bar was the cause of a LOT of problems with Live creating erratic self-growing windows that were a literal nightmare to fix. A menu bar band that does not match Live's model is the resize-growth failure I spent a long time on, represented in patch `0029`.

shibco (changes requested):
It's almost there!

scripts/ableton-live
ClickSentinel: Fixed in 1505387 — gated on fc-list :lang=ja returning something real before trusting fc-match's result.

scripts/ableton-live
ClickSentinel: Updated in 1505387 to explain why fc-list, not fc-match, is what actually detects 'no CJK font'.

patches/0054-win32u-fall-back-to-a-linked-font-for-missing-glyph.patch
ClickSentinel: Fixed in 9875172 — surrogate halves (0xD800–0xDFFF) get the same skip control characters already had.

scripts/ableton-live
ClickSentinel: Fixed in 1505387 — the symlink write now happens after the early-return check, and a stale symlink from a previous host CJK font choice gets pruned, same pattern as sync_ui_font's AbletonSans* pruning.

scripts/ableton-live
ClickSentinel: Fixed in 1505387 — off now clears any leftover SystemLink registration for Ableton Sans/Ableton Sans Small before returning.

patches/0054-win32u-fall-back-to-a-linked-font-for-missing-glyph.patch
ClickSentinel: Tested this at multiple DPI sizes with many different window resize patterns and could not reproduce 0029. Thanks for pointing this one out - that looked like an absolute PITA.

amenohi2:
I'll check out the Japanese interface later when I get a chance.

amenohi2:
I checked out the menus and everything looks fine. No mojibake or cut off text or anything like that.

This probably isn't a bug with the PR but with my environment instead, but font is looking pretty bad for me. Your screenshots look a lot better. Any ideas? I tried installing cjkfonts using winetricks.

image

Edit: Various regedits later, it's looking ok now. 
image

How I did it (on CachyOS/Arch):
Installed the `noto-fonts-cjk` then created `~/.config/fontconfig/fonts.conf` with the following:
```xml
<?xml version="1.0"?>
<!DOCTYPE fontconfig SYSTEM "urn:fontconfig:fonts.dtd">
<fontconfig>
  <alias>
    <family>sans-serif</family>
    <prefer>
      <family>Noto Sans</family>
      <family>Noto Sans CJK JP</family>
    </prefer>
  </alias>
  <alias>
    <family>serif</family>
    <prefer>
      <family>Noto Serif</family>
      <family>Noto Serif CJK JP</family>
    </prefer>
  </alias>
</fontconfig>
```

Also this reg for wine: https://gist.github.com/nogajun/6095ed8488ef9449e63dc1ce578ae55e

Then ran `fc-cache -f` and completely restarted wine (`wineserver -k` then reopened Ableton).

LGTM!

shibco:
Thanks for testing this @amenohi2 .

Wine font rendering is a real problem. I would be open to seeing if there were alternatives we could add into this fork to standardise it across distros.

ClickSentinel:
@amenohi2 Thanks again for your thorough testing. 

This PR will fall back to system-installed fonts, and it looks like the out-of-box one with your system was not very pretty.  I'll take this away and see what options are available.

PR #80, opened 2026-07-27
Make Ableton Link work out of the box

shibco:
This PR provides additional setup support for `ableton-linkd`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,10 @@
   as the X11 `PMinSize` hint, and the window manager stops the drag at
   the minimum.
 
+- The installer now configures Ableton Link networking during installation.
+  Pass `--no-link` to skip the privileged multicast-route and firewall step.
+  The NetworkManager hook now keeps the route on the current non-VPN default
+  interface.
 - Fixed a Live crash when closing WebView2 plugin editors (issue 52, Wine
   patch 0045). `RevokeDragDrop` now rejects windows owned by another process,
   matching `RegisterDragDrop`. Fix by Giang Nguyen. See
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,10 +11,12 @@
   as the X11 `PMinSize` hint, and the window manager stops the drag at
   the minimum.
 
-- The installer now configures Ableton Link networking during installation.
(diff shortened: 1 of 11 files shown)


ClickSentinel:
An overarching note I would make about this is that the install will now require `sudo` by default, which I think is something to handle carefully. I'm not sure what I would do differently with that information but possibly a README.md note or something to reflect that.

ClickSentinel:
`configure_link()` short-circuits if `~/.local/share/ableton-wine/link-configured` already exists - it never re-runs `setup-link.sh`, so it never compares/replaces the old hook file. Anyone who ran the old `setup-link.sh` before this PR keeps the old, VPN-unsafe hook indefinitely after upgrading via --update, unless they manually delete the marker or re-run setup-link.sh by hand.

I would recommend making the marker version-aware instead of bare existence check so a logic change forces exactly 1 re-run on upgrade, but steady-state updates skip cleanly. 

See PR #83 for all recommended changes.

ClickSentinel (changes requested):
See PR #83 for details

shibco:
Hey @ClickSentinel guess what!

Setup no longer needs `sudo` 
(in most cases) 

My original approach was overengineered: the Link SDK already handles mutlicast discovery and bypasses the routing table completely. 

So this 'v3' design defers more to Link than its own setup.

As far as I can see, there are two instances where `sudo` is required for networking:

1. opening a port (UDP `20808`)  when the host has an active firewall
2. removing the shitty old OG hook I made on upgrade, once. If that removal fails, setup exits without recording v3 complete, so the next update retries. Declined (--no-link) installs get the removal command printed instead of a silent leftover.

I closed #83 because it is merged into this PR! Thank you for those commits! Please review the whole diff. 

In particular, no matter what, if this fails:

```
sudo ip route del 224.0.0.0/4
~/.local/share/ableton-wine/ableton-linkd --probe 10
WINEPREFIX="$HOME/.wine-ableton" \
  "$HOME/.local/opt/wine-d2d1-nspa-11.13/bin/wine" tools/linkprobe.exe
```
Block block approval until we fix it. No-route operation is the whole premise of this change.

ClickSentinel (approved):

ClickSentinel (approved):

PR #83, opened 2026-07-28
Make Link opt-out sticky and version the setup marker

ClickSentinel:
PR for a combination of fixes into #80 

`configure_link()` treats the marker file (`~/.local/share/ableton-wine/link-configured`) as a simple "have we done this before?" flag. That causes two things:

1. Anyone who already ran the old `setup-link.sh` has that marker sitting there, so `--update` will never touch it again — which means they never actually get the new VPN-aware NetworkManager hook. They're stuck on the old, VPN-unsafe one unless they know to delete the marker or re-run the script themselves. Kind of ironic since they're the people most likely to hit the bug this branch fixes.
2. `--no-link` doesn't stick. If someone opts out on their first install and later runs `--update` without remembering to pass `--no-link` again, they get hit with a surprise `sudo` prompt on what should've been a quiet update.

fixes:

- `setup-link.sh` now carries a `LINK_SETUP_VERSION`, written into the marker next to the interface name. `configure_link()` only skips when the recorded version matches what's expected, so bumping it forces one clean re-run on affected machines and then goes back to being quiet.
- `--no-link` now writes a `declined` state into that same marker, so it's respected on later `--update` runs without having to repeat it every time. Added `--link` as the explicit opt-back-in (shows up in `--help`, wired through the arg parser and into `configure_link()`).

Ran `bash -n` + shellcheck on both files, and walked through all six marker states by hand (fresh install, rerun at the same version, rerun at a stale version, `--no-link`, a later `--update` without re-passing it, and `--link` overriding a decline) — behaves as expected in each case.

Only `scripts/setup-link.sh` and `scripts/setup-run-header.sh` are touched, nothing else.

### DIFF
--- a/scripts/setup-link.sh
+++ b/scripts/setup-link.sh
@@ -5,6 +5,10 @@
 # ~/.local/share/ableton-wine/ableton-linkd); this script never builds software.
 set -euo pipefail
 here="$(cd "$(dirname "$0")" && pwd)"
+# Bump whenever this script's system-level effects change (hook body, firewall
+# rule, unit handling): forces one re-run on hosts with a stale marker so an
+# existing install picks up the fix instead of silently keeping old behavior.
+LINK_SETUP_VERSION=2
 
 if pgrep -f "Ableton Live.*\.exe" >/dev/null 2>&1; then
     echo "!! Live is running: close it before changing Link networking" >&2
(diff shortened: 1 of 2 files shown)


shibco:
Before I look at this, I will do some additional research to see if there's a way to avoid `sudo` all together. Brbrbrbbrbrbrbrbrb

PR #94, opened 2026-07-29
Prefer GL present for top-level swapchain windows (issue 91)

### LINKED ISSUE #91: Present Storm with GPU Rendering Enabled
## What happened

With Live's GPU renderer enabled (PR #74, which removes `-_ForceGdiBackend`), every `Present()` of Live's main window takes wined3d's GDI fallback path instead of presenting from the GPU. Each present reads the whole backbuffer off the GPU into system memory, wraps it in a GDI bitmap, and pushes it to the X server as a full-window image — about 14 MB per frame at 2560x1350.

Idle costs nothing, since nothing is presenting. Any sustained UI activity turns it into a firehose, and moving the mouse across the window is enough:

| | display-server traffic | CPU |
| --- | --- | --- |
| GPU renderer as shipped | ~650 MB/s | >1 full core (95-120%) |
| with `PREFER_GL_PRESENT` set | ~0.4 MB/s | ~20% of a core |

Idle CPU over a 30 s sample: **70.6% of a core before, 16.4% after.**

This is likely a contributor to the "Live freezes / becomes unresponsive" reports (#61) — a process saturating a core and pushing 650 MB/s at XWayland just to track the mouse feels exactly like that from the outside. I'd stop short of calling it the cause without someone else reproducing.

## Root cause

wined3d already has the fast path: present straight from the GPU with `glXSwapBuffers`, no readback, no CPU copy. It's gated behind `WINED3D_SWAPCHAIN_PREFER_GL_PRESENT`, which exists and works — it was just never set for this case.

The only place that turned it on was the DirectComposition branch in `dlls/dxgi/factory.c`, for composition targets. Live's main window is a plain top-level swapchain window, so it never entered that branch and always took the GDI path.

The whole difference is one argument in `swapchain_gl_present()`:

```c
/* GDI path — drags the whole backbuffer off the GPU every frame */
wined3d_texture_load_location(back_buffer, 0, context, WINED3D_LOCATION_SYSMEM);

/* GL path — leaves it on the GPU and flips */
wined3d_texture_load_location(back_buffer, 0, context, back_buffer->resource.draw_binding);
```

The path change is directly observable, not inferred: wined3d's fire-once `FIXME("Using GDI present.")` in `swapchain_blit_gdi()` fires on a normal startup without the fix and doesn't fire at all with it. The expensive function stops being called rather than getting cheaper.

## Steps to reproduce

1. Use a prefix with the GPU renderer enabled — no `-_ForceGdiBackend` in `Options.txt`, which is what `setup-prefix.sh` step 5c produces.
2. Open Live with the main window at a normal size.
3. Move the mouse continuously across the window for a few seconds.
4. Watch `/proc/<pid>/stat` fields 14+15 for CPU, and the `wchar` field of `/proc/<pid>/io` for bytes written to the display server.

To confirm which path is in use:

```bash
WINEDEBUG=fixme+all,err+all ableton-live > log 2>&1
grep -c "Using GDI present" log      # 1 = GDI readback path, 0 = GL path
```

Two measurement notes. `scripts/ableton-live` sets `WINEDEBUG="${WINEDEBUG:--all}"`, so anything relying on default `err`/`fixme` output silently produces nothing unless you pass `WINEDEBUG` explicitly. And `wchar` counts the debug log's own writes, so leave tracing off when measuring bandwidth.

## Notes for a fix

Setting the flag at swapchain creation for top-level device windows works, but it has to exclude both `WS_CHILD` and `WS_POPUP`, or secondary windows (Settings, the authorisation dialog, right-click menus) paint black until you interact with them. Two things make the window test less obvious than it looks:

- `WS_POPUP` and `WS_CAPTION` coexist — Live's Settings window is framed and still `WS_POPUP` (style `96c80000`).
- `CreateSwapChainForComposition()` creates its own `WS_CHILD` window parented to the desktop, so `GetAncestor(GA_PARENT)` returns the desktop and a parent-only test wrongly calls it top-level.

## System summary

Measured on AMD Navi 31 (RX 7900 XT/XTX), COSMIC / Wayland via XWayland, Live 12.4.3 Suite, runtime `2026.07.23.1`, Wine base `d2d1-dcomp-11.13` (`5c23dd1c`). Single machine and compositor so far — worth reproducing on Intel/NVIDIA or a non-Wayland session before treating the numbers as universal.

shibco:
WIP PR for patch `0055` based on investigations by @ClickSentinel.

So from what I understand, `wined3d` displays every frame of Live's main window through GDI. 

So this bug copies the finished frame from the graphics card into main memory and sent it to the display server as a full-window image, which is like 600mb per second during continuous UI activity. 

This patch sets `WINED3D_SWAPCHAIN_PREFER_GL_PRESENT` when a top-level window's frame buffers are created, so `wined3d` shows each frame directly from the graphics card with `glXSwapBuffers`. 

`WS_CHILD` and `WS_POPUP` windows keep the GDI path. `WINE_DISABLE_GL_PRESENT=1` restores the GDI path for every window.

Please test!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,13 @@
   (Wine patch 0053). winex11 now exports the `WM_GETMINMAXINFO` minimum
   as the X11 `PMinSize` hint, and the window manager stops the drag at
   the minimum.
+- Fixed high CPU use and display traffic with the GPU renderer (issue 91,
+  Wine patch 0055). Wine copied every finished frame of Live's main window
+  from the graphics card back into main memory and sent it to the display
+  server as a full image, about 650 MB per second during continuous UI
+  activity such as mouse movement. Wine now shows finished frames directly
+  from the graphics card. Set `WINE_DISABLE_GL_PRESENT=1` to restore the
+  previous behaviour. Diagnosis and measurements by Lucas Gillingham.
 
 - The installer now configures Ableton Link during installation. Setup no
   longer adds a multicast route or NetworkManager hook: the Link SDK selects
--- a/notes/ABLETON-WINE-GPU-RENDERER.md
+++ b/notes/ABLETON-WINE-GPU-RENDERER.md
@@ -59,11 +59,48 @@ Checks that still need a pass after longer real-world use: file dialogs
 under sustained work, plugin editor open/close (JUCE, SWAM), and both
 panes across scale factors other than 200%.
 
+## Present path (added 2026-07-29, issue 91)
+
+A present is the step where Live hands a finished frame to Wine for
+display. With the GPU renderer on, Wine handled every present of Live's
+main window on its GDI path: it copied the finished frame from the
+graphics card into main memory and sent it to the display server as a
+full-window image, about 14 MB per frame at 2560x1350. An idle window
+presents nothing, so the idle figures above stay correct. Continuous UI
+activity, including mouse movement over the window, produced about
+650 MB per second of display-server traffic and used more than one CPU
+core. Lucas Gillingham (ClickSentinel) reported and measured this in
+issue 91.
+
+Wine patch 0055 marks the main window's frame buffers at creation with
+`WINED3D_SWAPCHAIN_PREFER_GL_PRESENT`. Wine then shows each finished
+frame directly from the graphics card with `glXSwapBuffers` and skips
+the copy. The patch applies this to top-level windows only. Windows
+with the `WS_CHILD` style (composition targets, embedded plugin
+editors) keep the GDI path because they have no X11 window of their
+own. Windows with the `WS_POPUP` style (Settings, the authorisation
+dialog, context menus) also keep the GDI path because they show black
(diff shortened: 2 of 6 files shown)


ClickSentinel:
Just one. 

`GetEnvironmentVariableW` returns 0 only when the variable is unset, so the check is presence-based: `WINE_DISABLE_GL_PRESENT=0` disables GL present just as `=1` does. The CHANGELOG and the note both document `=1`, so someone who sets `=0` to keep the new behaviour gets the old one instead. Either document it as "set to any value" or read the value:

    WCHAR v[8];
    BOOL disabled = GetEnvironmentVariableW(L"WINE_DISABLE_GL_PRESENT", v, ARRAY_SIZE(v)) && v[0] != '0';

Presence-based is a perfectly normal Wine convention — it's only the `=1` in the docs that makes it surprising.

shibco:
Thanks, @ClickSentinel! fixed in `f0bd6a9`

ClickSentinel (approved):

PR #97, opened 2026-07-29
Identify Intel GPUs properly

shibco:
This PR makes Live's GPU renderer available again on Intel graphics newer than 2019 (issue 84, Wine patch 0057). Wine reported every Intel GPU from Ice Lake through Lunar Lake, and the Arc A-series cards, as "Intel(R) HD Graphics 4000", a 2012 device that Live refuses to use, so the Preferences dialog greyed out "Enable GPU Renderer". Wine now reports the real device names. 

Thank you to @stickyfran for being persistent with this!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,14 @@
 # Changelog
 
+## Unreleased
+
+- Live's GPU renderer is available again on Intel graphics newer than 2019
+  (issue 84, Wine patch 0057). Wine reported every Intel GPU from Ice Lake
+  through Lunar Lake, and the Arc A-series cards, as "Intel(R) HD Graphics
+  4000", a 2012 device that Live refuses to use, so the Preferences dialog
+  greyed out "Enable GPU Renderer". Wine now reports the real device names.
+  Reported by stickyfran.
+
 ## 2026.07.29.1
 
 - Live now uses its GPU renderer. Prefix setup removes the legacy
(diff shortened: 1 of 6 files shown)


shibco:
@ClickSentinel I think this one got superseded by your GPU work, right?

ClickSentinel:
@shibco  it was supposed to but may need more work as per most recent comment on that PR. We may need to build a diagnostic script or work with @stickyfran

shibco:
The issue is that the new work doesn't satisfy the deny list. I am going to merge this one over the top of the work already done, but this may crop up again on a case by case basis and need some additional thinking.

shibco:
Edit: to clarify, @ClickSentinel its that this is literally an annoying Live problem, the work on your more recent PR will fix 90% of the cases of this cropping up

shibco:
I rebased onto current main, keeping #105. The patch is unchanged.

@ClickSentinel's work on #105 builds the description from the driver's renderer string, and for a Meteor Lake iGPU that comes out as "Intel(R) Arc(tm) Graphics", the device's real Windows name. @stickyfran's report on #105 shows Live rejects that name too, which matches Ableton's known Arc trouble on Windows. The table entry this PR ensures the "(MTL)" suffix is what gets past Live's exact-name check. 

The two patches complement each other: the table covers the denylisted Intel names, `0061` covers everything else.

ClickSentinel:
Ok sounds good. Part of the reason this broke out into a separate PR was that I found some issues with IDing newer Intel GPUs - when I have service again I can pull up those notes.

shibco:
@stickyfran, please try now!

PR #98, opened 2026-07-29
launcher: keep the GDI present path; 0055's GL path misplaces the frame

Version33:
A patch claude wanted to put in another PR
didn't see why not, so here's that to look at

just some docs and a patch that solves an issue I was having with Live rendering with a massive black bar on the upper third of the screen; fis is just an env var. If this is just a me issue and making `WINE_DISABLE_GL_PRESENT` 1 by default is not a good idea, go ahead and close this. Or if you don't like the docs/comments, lmk and I can clean those up

claude says:
```
Patch 0055 (PR #94, issue 91) prefers wined3d's GL present path for top-level swapchain windows. On niri -- scrollable tiling, Wayland, XWayland, AMD/radv -- that path presents Live's frame roughly 476px low: a black band fills the top of the client area, the bottom of the mixer is clipped off-screen, and the whole UI looks displaced. Input hit-testing stays on the real layout, so clicks land where a control should be rather than where it is drawn.

Correct geometry with misplaced pixels puts the fault in the present blit. Measured on a 3840x2160 output at scale 1.25 (LogPixels 120): the X11 window was 3840x2045 and its client child 3840x2017 at +0+28, and WM_NORMAL_HINTS carried 0053's PMinSize 1208x999 -- all correct. WINE_DISABLE_GL_PRESENT=1 renders correctly and logs "Using GDI present" once; the same build without it never logs it and misplaces the frame.

Default WINE_DISABLE_GL_PRESENT to 1 in both launchers, in the same ${VAR:-default} form as the other wined3d knobs there, so the stack keeps the copy path and its bandwidth cost until the direct path is fixed. WINE_DISABLE_GL_PRESENT=0 opts back in; 0055 reads the value, so 0 means the GL path, not "disabled". The patch series is untouched.

Likely mechanism, unconfirmed: a tiling compositor resizes the window right after mapping and the direct path bottom-aligns the pre-resize buffer in the new drawable (GL's origin is bottom-left, X11's is top-left). That would make every tiling compositor affected and every floating one clean, matching the COSMIC measurements in notes/ABLETON-WINE-GPU-RENDERER.md. Untested on sway, Hyprland and i3, so this is a blanket default rather than a per-compositor gate.
```

### DIFF
--- a/notes/ABLETON-WINE-GPU-RENDERER.md
+++ b/notes/ABLETON-WINE-GPU-RENDERER.md
@@ -96,6 +96,35 @@ These measurements come from one machine (AMD Navi 31, COSMIC/Wayland
 via XWayland). Confirmation on Intel or NVIDIA hardware and on a
 non-Wayland session is still open.
 
+### The direct path misplaces the frame on niri (2026-07-29)
+
+On niri (scrollable-tiling Wayland compositor, XWayland, AMD/radv) the
+direct path presents Live's frame at the wrong vertical origin. The
+image sits about 476 px low: a black band fills the top of the client
+area, the bottom of the mixer is clipped off-screen, and the whole UI
+looks displaced. Input hit-testing stays on the real layout, so
+clicking works where a control *should* be, not where it is drawn.
+That split — correct geometry, misplaced pixels — puts the fault in the
+present blit, not in Live's layout or the DPI policy.
+
+Measured on a 3840x2160 output at scale 1.25 (`LogPixels` 120). The X11
+window was 3840x2045 and its client child 3840x2017 at +0+28, both
+correct. `WINE_DISABLE_GL_PRESENT=1` renders correctly and logs `Using
+GDI present` once; the same build without it logs nothing and misplaces
+the frame.
+
+Unconfirmed, but the likely mechanism is that a tiling compositor
+resizes the window immediately after mapping, and the direct path
+bottom-aligns the pre-resize buffer in the new drawable (GL's origin is
+bottom-left, X11's is top-left). If that holds, every tiling
+compositor is affected and every floating one is not, which matches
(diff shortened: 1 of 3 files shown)


Immabed:
fixes #100

shibco:
First, @Immabed @NoskyD and @Version33, the immediate fix is to load Ableton Live with this command: 

```bash
WINE_DISABLE_GL_PRESENT=1 ableton-live
```

I tested this fix and despite it being very small it is also deceptively simple.

So.

There are two ways to draw Live's interface on screen. 

Option A - Copy the picture "by hand" every frame, where we favour accuracy and ensure it always looks right but wastes a lot of the computer's effort. 

Option B - Send the picture straight to the graphics card. 

But because nothing in Windows and Wine life is simple, the fast, cool and very performant Option B has a drawback: a Windows window changes size right after it opens, which some desktops handle automatically, and which Live's HiDPI setting also does. In some circumstances, the interface never fully updates and the GPU thinks the window is a different size, and you get a magnificant void between your set and the top of the Live window.

This PR as it stands is a little bit of a monkey-paw wish trade: it will immediately fix this black void issue for everyone, everywhere, and everyone gets slugged with a huge performance drop that is noticeable depending on the number of CPU cores you have, etc etc. OR some users keep the black void unless they manually disable the GDI prsent path with `WINE_DISABLE_GL_PRESENT=1`

There are a couple of things we can try. I'll keep this PR open until I know exactly what to do.

NoskyD:
Yup, that works! (Yet you have to sacrifice some of the performance, it lags sometimes... I thought the #61 would come back, but luckily no)

shibco:
Thanks for confirming @NoskyD! We got a fix over at #102 that @ClickSentinel is gonna test when he can

shibco:
This should be fixed now by #102. Closing.

Issue #100, opened 2026-07-30
Live's UI slides down with HiDPI setting on (Laptop GTX 1050)


NoskyD:
## What happened

I'm running Live 12 Suite on CachyOS with KDE Plasma.
The GPU is a laptop version of NVIDIA GTX 1050, the driver is `nvidia-580xx-dkms 580.173.02-1`.

When the 'Enable HiDPI Mode' setting is turned on, Live's interface renders slightly to the bottom with a huge black gap between the window's top bar and (which Live has).

It doesn't depend on the GPU setting:

<img width="1919" height="1080" alt="Image" src="https://github.com/user-attachments/assets/b47d85d0-2901-4218-ad52-cd519975c806" />
<img width="1919" height="1080" alt="Image" src="https://github.com/user-attachments/assets/65e0e295-ae53-4287-831b-bb848e644c80" />

## Steps to reproduce

1. Run Ableton Live on him with 'Enable HiDPI Mode' setting turned on.
2. Notice the UI sliding down and a black gap at the top.

## System summary

Run the profiler on the affected machine from the root of this repository. It
prints a redacted system summary and copies it when a supported clipboard tool
is available.

If the same setup works on macOS or Windows, add that profiler's output for
comparison.

- Linux: `./beta/scripts/ableton-linux-profiler.sh`
- macOS: `./beta/scripts/ableton-macos-profiler.sh`
- Windows: `powershell -ExecutionPolicy Bypass -File beta\scripts\ableton-windows-profiler.ps1`

Review the summary, then paste it here:
*Didn't find the script...*

Immabed:
I'll add that this is a regression. On an earlier version of ableton-linux with wine 11.11, Ableton rendered fine. Not sure which of many changes could have caused it.

shibco:
Fixed in #102

PR #102, opened 2026-07-30
Patch 0058: GDI present for frames whose client rect disagrees (issue 100)

shibco:
The GL present path (0055) can land Live's frame too low: black band on top, bottom clipped (#100; same signature as the niri report on #98).

Cause candidate: the destination rect is captured on Live's thread in the window's DPI context (0023), but the blit's y-flip re-queries the client rect later on wined3d's CS thread in that thread's DPI context. If the two disagree (thread DPI mismatch, or a resize in flight), GL's bottom-left origin lands the frame low by the difference.

Patch 0058 compares the two rects on every frame. Match: GL present as before. Mismatch: that frame takes the GDI path, which is correct under the same conditions. Affected setups end up on the GDI path full-time, as if WINE_DISABLE_GL_PRESENT=1; healthy setups keep the GL path. A fire-once FIXME and a rate-limited TRACE log both rects for diagnosis.

Build and 8/8 audit pass. Runtime verification on an affected setup (niri, or the #100 machine) is pending. If the bar survives there without the gate firing, the disagreement is driver-side and the follow-up is a DPI-consistent query in translate_drawable_coords.

If this verifies, this supersedes the proposed flag switch in #98.

### DIFF
--- a/notes/ABLETON-WINE-GPU-RENDERER.md
+++ b/notes/ABLETON-WINE-GPU-RENDERER.md
@@ -127,7 +127,48 @@ rate-limited TRACE record both rects and the backbuffer size, so an
 affected machine can show which side lies, toward a root fix in
 `translate_drawable_coords` itself.
 
-Runtime verification on an affected compositor is pending.
+Runtime verification is below: the gate is correct, and the
+disagreement it detects has a root cause worth fixing.
+
+### The disagreement is a DPI-context gap; patch 0059 closes it (2026-07-30)
+
+The trigger is fractional display scaling, not a compositor, a driver
+or a window manager. It reproduces on AMD Navi 31 under COSMIC/Wayland
+— a setup with no symptom at 100% — by putting the prefix at 125%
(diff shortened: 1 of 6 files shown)


ClickSentinel:
The cost of this fix is that at 125% scaling, cpu usage for _drawing_ the interface increases ~3 fold.

| Configuration | Wine scale | Black bar | Live CPU (mean / peak) | Present path |
|---|---|---|---|---|
| main today — 0055, no 0058 | 125% (LogPixels 120) | **Yes** | 29.8% / 35.5% | GL (fast) |
| PR #102 — 0058 applied | 125% (LogPixels 120) | No | 98.9% / 117.5% | GDI (slow) |
| PR #102 — 0058 applied | 100% (LogPixels 96) | No | 26.6% / 29.4% | GL (fast) |

Workaround for this bug is set DPI scale at launch to 100% and use the zoom feature in Ableton.

Personally, I'd like some time to investigate a fix that doesn't spike the CPU. If this is urgent, ship it, issue, fix later. 

Edit: fixed in #107

PR #103, opened 2026-07-30
Fix mp3 import: winegstreamer

### LINKED ISSUE #44: .mp3 files fail to load (can cause export failures)
Live fails to load any .mp3 file, showing the error "The file could not be read. It may be corrupt or not licensed."

Image

When double clicking on the file, Live shows the error "This sample contains no audio data:"

Image

When exporting audio as a .mp3, if 'Create Analysis File' is enabled, then when Live finishes rendering the audio, it will show "This sample contains no audio data" error and the encoded .mp3 file will be discarded. If the analysis option is disabled, the export finishes successfully.

---
[Discord thread](https://discord.com/channels/1307779762189504602/1530636023162343748/1530636023162343748)

ClickSentinel:
## Summary

`Containerfile` never installed `libgstreamer1.0-dev`/`libgstreamer-plugins-base1.0-dev`. Wine's `configure` treats GStreamer support as opt-out (`--without-gstreamer`), not opt-in, so without those packages it silently dropped the `winegstreamer` module — no build failure, just a notice buried in the configure log.

Without `winegstreamer.dll`/`.so`, `mfplat`'s Media Foundation source resolver has no backend to hand media files to, so mp3 import failed on every build. The exact error depends on prefix history: `MF_E_UNSUPPORTED_BYTESTREAM_TYPE` on a clean prefix, or `E_OUTOFMEMORY` out of `wg_parser_create` on a prefix carrying a stale leftover `winegstreamer.dll` from an older build with no matching backend. Same root cause either way.

This adds the two dev packages to `Containerfile`, plus build-audit gates in `scripts/container-build.sh` and `scripts/install.sh` that fail loudly if `winegstreamer.so` doesn't actually get produced — the same protection the build already has for `winealsa` and `ntsync` silently dropping.

Fixes #44.

## Test plan

- [x] Build audit passes (76 checks); packaged tarball now contains `lib/wine/x86_64-unix/winegstreamer.so` and `winegstreamer.dll` (both previously absent).
- [x] Installed over a real `~/.wine-ableton` and confirmed mp3 import works.

### DIFF
--- a/Containerfile
+++ b/Containerfile
@@ -56,6 +56,10 @@ RUN apt-get update && apt-get install -y --no-install-recommends \
       # driver. PipeASIO builds against the vendored PipeWire SDK below, not a
       # jammy package (jammy's 0.3.48 predates the thread-utils API it needs).
       libasound2-dev libpulse-dev \
+      # media import: without these, configure silently drops winegstreamer
+      # and mp3/mp4/wma import just fails (issue #44). Actual codec plugins
+      # still come from the user's host GStreamer install at runtime.
+      libgstreamer1.0-dev libgstreamer-plugins-base1.0-dev \
       # TLS (Live online auth / pack downloads), USB display bridge, XDG portal
       libgnutls28-dev libusb-1.0-0-dev libudev-dev libdbus-1-dev \
  && rm -rf /var/lib/apt/lists/* \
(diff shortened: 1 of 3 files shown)


shibco:
This is good to merge with two little updates: 

1. add `winegstreamer` entries to `build-audit.sh`'s structural invariants
2. add gstreamer to the list of requirements in README.md

Approving because these are tiny changes and we can merge.

PR #105, opened 2026-07-30
Patch 0061: wined3d: describe unlisted GPUs from the driver's own name

### LINKED ISSUE #84: (  2026.07.29.1 2026.07.29.1 )No HW Acceleration on Intel Arc Asus Laptop S5506MA
Image

UI seems laggy on this laptop. It's an Intel 155h.

[SYSTEM]
PRETTY_NAME="CachyOS"
Linux 7.2.0-rc4-1-cachyos-rc x86_64
CPU(s):                                   22
               total       usado       libre  compartido   búf/caché  disponible
Mem:            15Gi        11Gi       1,0Gi       1,8Gi       4,0Gi       3,4Gi
sys_vendor=ASUSTeK COMPUTER INC.
product_name=ASUS Vivobook S 15 S5506MA_S5506MA

[GRAPHICS]
OpenGL vendor string: Intel
OpenGL renderer string: Mesa Intel(R) Arc(tm) Graphics (MTL)
OpenGL core profile version string: 4.6 (Core Profile) Mesa 26.2.0-devel (git-584a0997c8)
OpenGL version string: 4.6 (Compatibility Profile) Mesa 26.2.0-devel (git-584a0997c8)
desktop=KDE
session=wayland

[AUDIO]
pipewire=active
pipewire-pulse=active
wireplumber=active
pipewire
Compiled with libpipewire 1.6.8
Linked with libpipewire 1.6.8
wireplumber
Compiled with libwireplumber 0.5.15
Linked with libwireplumber 0.5.15
Found "settings" metadata 31
update: id:0 key:'log.level' value:'2' type:''
update: id:0 key:'clock.rate' value:'48000' type:''
update: id:0 key:'clock.allowed-rates' value:'[ 48000 ]' type:''
update: id:0 key:'clock.quantum' value:'1024' type:''
update: id:0 key:'clock.min-quantum' value:'32' type:''
update: id:0 key:'clock.max-quantum' value:'2048' type:''
update: id:0 key:'clock.force-quantum' value:'512' type:''
update: id:0 key:'clock.force-rate' value:'48000' type:''
**** List of PLAYBACK Hardware Devices ****
card 0: Audio [AB13X USB Audio], device 0: USB Audio [USB Audio]
  Subdevices: 0/1
  Subdevice #0: subdevice #0
card 1: sofhdadsp [sof-hda-dsp], device 0: HDA Analog (*) []
  Subdevices: 1/1
  Subdevice #0: subdevice #0
card 1: sofhdadsp [sof-hda-dsp], device 3: HDMI1 (*) [HDMI 1]
  Subdevices: 1/1
  Subdevice #0: subdevice #0
card 1: sofhdadsp [sof-hda-dsp], device 4: HDMI2 (*) [HDMI 2]
  Subdevices: 1/1
  Subdevice #0: subdevice #0
card 1: sofhdadsp [sof-hda-dsp], device 5: HDMI3 (*) [HDMI 3]
  Subdevices: 1/1
  Subdevice #0: subdevice #0
card 1: sofhdadsp [sof-hda-dsp], device 31: Deepbuffer HDA Analog (*) []
  Subdevices: 1/1
  Subdevice #0: subdevice #0

[MIDI]
Dir Device    Name

ClickSentinel:
On any Intel GPU newer than about 2019, Live greys out "Enable GPU Renderer" and says `Intel(R) HD Graphics 4000: Unexplained slow UI at zoom-level 100% and/or crashes`. That's issue 84, reported on a Core Ultra 7 155H, which is a 2024 part. Patch 0035 ran into the same thing with Battlemage and fixed it by adding one row to the device table.

Live's check goes by name, not by what the card can do, and wined3d is telling it the wrong name. wined3d only names a card it can find in `gpu_description_table`. When the lookup misses it guesses a device from the feature level instead, and for Intel that guess is `CARD_INTEL_IVBD`, whose table entry is literally `"Intel(R) HD Graphics 4000"`. So every Intel GPU that isn't in the table ends up reporting the exact string Live refuses.

This isn't only an Intel problem. My 7900 XT isn't in the table either:

```text
query_gpu_description  Card reports vendor PCI ID 0x1002, device PCI ID 0x744c, 0x500000000 bytes of video memory.
wined3d_guess_card     Unrecognized renderer "AMD Radeon RX 7900 XT (radeonsi, navi31, ACO, ...)", falling back to default.
wined3d_adapter_init_gl_caps  Guessed device PCI ID 0x68d8.
```

`0x68d8` is "ATI Radeon HD 5600 Series", a 2010 card with 512 MB. Both vendors fall back to something ancient. AMD's just happens to land on a name Live accepts and Intel's doesn't. You can force it either way on hardware that's otherwise fine:

```bash
WINE_D3D_CONFIG=VideoPciVendorID=0x8086,VideoPciDeviceID=0x0162 ableton-live   # greyed out
WINE_D3D_CONFIG=VideoPciVendorID=0x8086,VideoPciDeviceID=0x3e9b ableton-live   # available
```

Same machine, same session, one PCI ID apart.

## The fix

By the time the lookup fails, wined3d already has everything it needs. `GLX_MESA_query_renderer` just gave it the real PCI IDs and the video memory, and `GL_RENDERER` has the device name sitting right there. The only thing that failed was the table lookup. So build the description out of what the driver reported instead of throwing it away and guessing.

The Vulkan back end already does this, using `VkPhysicalDeviceProperties.deviceName`. This is the same thing for OpenGL. Mesa's `Mesa` prefix and the parenthesised driver detail get trimmed off, since no Windows driver reports either of those.

Two situations still go down the old path. Mesa uses `0xffffffff` for a renderer that isn't a PCI device at all, and `PCI_VENDOR_NONE` is the 16-bit `0xffff`, so an obvious-looking comparison ends up inventing a graphics card for llvmpipe. The three renderer attributes are also queried separately, so if the memory query fails while the device ID succeeds, you'd end up reporting a card with zero bytes of VRAM. The old fallback at least has the table's rough guess. Both cases decline and fall through.

Beyond that nothing moves. A device that's already in the table never gets here, `WINE_D3D_CONFIG` still overrides, and the driver class and VRAM come out identical. The only thing this can do is replace a wrong answer with a right one.

## Why not just add the device IDs

Because the table can't keep up. I diffed `pci.ids` against it and came up with roughly 95 missing modern Intel IDs: 28 Arc A-series, 15 Raptor Lake, 13 Panther Lake, 13 Battlemage, 9 Alder Lake, 8 Comet Lake, plus Arrow Lake-H. These aren't obscure either. `0x46b0` through `0x46ba` are normal 12th-gen Iris Xe laptops and `0xa781` to `0xa788` are Raptor Lake-S desktops. Panther Lake is already shipping, so that's another round due. And every row has to survive each Wine base bump.

## Testing

| stack | reported | result |
| --- | --- | --- |
| AMD Navi 31 | `1002:744c` | `"AMD Radeon RX 7900 XT"` (was "ATI Radeon HD 5600 Series") |
| virgl guest | `1af4:1050` | `"virgl"` |
| llvmpipe | `0xffffffff` | declined, unchanged |

That's not just the reported string. This machine's GPU isn't in the table, so Live is currently running on a synthesised description, with 156 D3D device and swapchain creations in the trace. I checked override precedence both in the trace and by watching the toggle grey out again. Build is clean and the audit passes 77/77.

What I couldn't test is the machine in issue 84. Two things are still guesses: that its device ID is `0x7d55` (I got that from its Mesa renderer string, not from `lspci`), and that Live doesn't also blacklist `"Intel(R) Arc(tm) Graphics"`. One traced launch from whoever reported it would settle both:

```bash
WINEDEBUG=+d3d ableton-live 2>&1 | grep -iE "Card reports|description table|HD Graphics"
lspci -nn | grep -i vga
```

---

Two bits of series bookkeeping come along with this, both currently wrong on main. `BASE.txt` still says 52 files through 0055, but 0056 has been shipping since #95 and never got counted or documented. And 0056 was put in `build-audit.sh`'s `STAMP_ONLY` list using the four-field `FINGERPRINTS` format, which means the parser read the whole line as a free-text reason and never opened `dxgi.dll` to check for it. It's been reporting `via stack stamp (ascii|lib/wine/...)` and verifying nothing this whole time. That one's my fault from #95. It does a real fingerprint check now.

This is probably worth sending upstream as well, since a 7900 XT reporting as a 2010 card affects anything that sizes buffers or picks quality presets off the GPU name, on any vendor. One thing a Wine reviewer would likely want changed: `guess_display_driver()` is copied from `adapter_vk.c` here, where it's static, and should really be shared instead.


ClickSentinel:
@stickyfran - I can't verify this directly with an intel GPU but it _should_ resolve the issue you were having. 

bonus - it will also use latest drivers for other undefined GPUs, like my 7900 XT. Would you mind testing against this build?

ClickSentinel:
Also - a bit of a reach but @amenohi2 you are the only person who has been happy to build who ALSO has an NVIDIA GPU. No obligation at all, but if you would be willing to test GPU rendering is still functional on your machine with this build, it would be incredibly useful. <3

stickyfran:
Patch from PR#97 fixed my issue
had to build without this patch cause this patch broke again gpu recognition :(

shibco:
@stickyfran I see what's happened here. Good news is that #97 can be rebased on top of this. Doing that now.

PR #108, opened 2026-07-30
Patch 0060: implement IFileOperation::DeleteItem

### LINKED ISSUE #45: Live fails to delete temporary files "for an unknown reason"
When closing an unsaved Live set with temporary files, if you click Delete, Live shows the error "For an unknown reason, Live could not delete the temporary files".

Image

---
[Discord thread](https://discord.com/channels/1307779762189504602/1530636048600662286/1530636048600662286)

trendwhore:
## Summary

- Add Wine patch `0060` implementing `IFileOperation::DeleteItem`.
- Fix Live's shared deletion path used by both browser deletion and backup pruning.
- Preserve recycle-bin behavior through the existing copy engine.

## Cause

Wine's `IFileOperation::DeleteItem` implementation is a stub that returns `E_NOTIMPL`. Ableton Live uses this function for both deleting files from its browser and pruning old project backups, causing both operations to fail:

```text
error: failed to move to trash "<project>/Backup/<old-backup>.als": io error
info: Exception: "<old-backup>.als" cannot be deleted.
```

## Validation

- The API reproduction changes from `E_NOTIMPL` to success.
- Live properly prunes down to the most recent 10 backups on default settings.

### DIFF
--- a/patches/0060-shell32-implement-IFileOperation-DeleteItem.patch
+++ b/patches/0060-shell32-implement-IFileOperation-DeleteItem.patch
@@ -51,7 +51,7 @@ index 16074d5..8ce6702 100644
  static void set_file_operation_progress(struct file_operation *operation, unsigned int total, unsigned int sofar)
  {
      operation->progress_total = total;
-@@ -1982,6 +2005,42 @@ static HRESULT copy_engine_move(struct file_operation *operation, struct copy_en
+@@ -1982,6 +2005,44 @@ static HRESULT copy_engine_move(struct file_operation *operation, struct copy_en
      return COPYENGINE_S_NOT_HANDLED;
  }
  
@@ -62,6 +62,8 @@ index 16074d5..8ce6702 100644
 +    unsigned int len;
 +    int result;
 +    HRESULT hr;
++
++    TRACE("IFileOperation DeleteItem via SHFileOperation.\n");
 +
 +    if (FAILED((hr = SHGetNameFromIDList(op->item_pidl, SIGDN_FILESYSPATH, &path))))
 +        return hr;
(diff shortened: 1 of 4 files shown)


ClickSentinel:
Thanks again for great work! Tested this and found no issues with the code. Since there are many PRs right now, our BASE.txt is a little all over the place. After that settles, we can get an entry added for this.

Approved from my side, awaiting Shibco's go-ahead. Then we can bring it in after other PRs in the queue! Fantastic!

shibco:
Thanks, @trendwhore!!!!
 
A couple tiny things before merge:

- [x] Live's browser allows deleting several files at once. If that goes through the plural `DeleteItems` call, it's still broken. Please test multi-select delete, and if it fails, cover `DeleteItems` too while you're in there.
- [ ] Add a unique trace message inside the delete path and register it as a fingerprint in build-audit.sh, instead of STAMP_ONLY. That lets the audit prove the fix is actually in a shipped build.
- [ ] Please include the API repro you used, in `tools/` so we can add it to our regression tests.

Notes for myself and @ClickSentinel -
- The audit note for `0057` is probably wrong and we should double check this before releasing
- This fixes #45

ClickSentinel:
@shibco multi select delete does actually work. I also found that in my testing but `DeleteItem` doesn't delete anything when called. It queues an entry via `add_operation`. The actual work happens in `PerformOperations`, which loops the whole queue: `LIST_FOR_EACH_ENTRY(op, &operation->ops, struct copy_engine_operation, entry)` Tested this with success on my side. The other 2 points are valid.

trendwhore:
Updated.

- Single and multiple deletes work with the current implementation.
- Added a unique `shell32` trace in the delete execution path and moved Patch 0060 from `STAMP_ONLY` to a compiled-binary `FINGERPRINTS` check.
- Added `tools/ifileoperation-delete.c`, a standalone API repro using Live-style `FOFX_RECYCLEONDELETE`. It succeeds only when `DeleteItem` and `PerformOperations` succeed, the operation is not aborted, and the original path is gone.
- Full rebuild and build audit pass.

shibco:
Accepted and merged! Thank you @trendwhore !!!

PR #117, opened 2026-08-02
Add Push 3 support.

shibco:
With help from @Version33, this patch extends the Ableton USB <-> Linux host instrument bridge to support the tethered Ableton Push 3, released exactly three years ago on 2 August, 2023. (Happy birthday Push 3!) 

Push3.exe imports four functions the Push 2 bridge does not export: `libusb_bulk_transfer` (forwarded to the host as one blocking call), `libusb_strerror` (standard message strings), and two hotplug functions.  As such, the new patch for the bridge enables it to accept interfaces 0 through 255 so the helper can reach the Push 3's `xPort` interface alongside the display.

`setup-prefix.sh`  deploys the builtin to `Push3.exe`, matching the Push 2 override. `tools/push3usb.c` is used to probe a real Push 3, and it is based off `push2usb.c`. 

As I do not have a Push 3, this PR can only be merged when someone (@Version33?) demonstrates the compatibility witha real device.

### DIFF
--- a/notes/ABLETON-WINE-PUSH2-DISPLAY.md
+++ b/notes/ABLETON-WINE-PUSH2-DISPLAY.md
@@ -58,7 +58,9 @@ HKCU\Software\Wine\AppDefaults\Push2DisplayProcess.exe\DllOverrides
     libusb-1.0 = builtin
 ```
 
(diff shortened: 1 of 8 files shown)


shibco:
Superseded by #254

PR #118, opened 2026-08-02
Performance moonshot

shibco:
image

This is part one of a three part optimisaiton process I've been working on since the initial release to stabilise audio and aggressively chase performance improvements. The goal here is to get this project running faster than the same version of Live running on Windows or macOS.

Note that the OSC control device will likely be replaced by this project with permission from its original authors after an evaluation: https://www.reddit.com/r/ableton/comments/1vcmehc/


shibco:
@ClickSentinel, let's discuss more about the opportunities to bring your full testing workflow into this branch as well?

shibco:
Forgot to add, this may supersede #49 and #73

patches/pipeasio/0004-accept-any-buffer-size-in-range-log-every-adjustment.patch
```diff
@@ -0,0 +1,68 @@
...
++                PIPEASIO_MIN_BUFFER_SIZE, PIPEASIO_MAX_BUFFER_SIZE,
++                c->buffer_size);
++    }
+ 
+     if (c->sample_rate < 0)
+         c->sample_rate = 0;
+diff --git a/src/asio.c b/src/asio.c
+--- a/src/asio.c
++++ b/src/asio.c
+@@ -1977,13 +1977,27 @@
```
ClickSentinel:
```suggestion
@@ -1289,10 +1289,11 @@
         TRACE("Buffersize fixed at %d\n", (int)This->host_current_buffersize);
     }
     else
-    { /* fail if not a power of two and if out of range */
-        if (!(bufferSize > 0 && !(bufferSize & (bufferSize - 1))
-              && bufferSize >= PIPEASIO_MINIMUM_BUFFERSIZE
-              && bufferSize <= PIPEASIO_MAXIMUM_BUFFERSIZE))
+    { /* Any size in range is accepted here too: Live builds non-power-of-two
+       * buffers, and 0005 adopts whatever quantum the graph runs (issue #49
+       * saw 192).  Otherwise adoption offers a size CreateBuffers rejects. */
+        if (!(bufferSize >= PIPEASIO_MINIMUM_BUFFERSIZE
+              && bufferSize <= PIPEASIO_MAXIMUM_BUFFERSIZE))
         {
             WARN("Invalid buffersize %d requested\n", (int)bufferSize);
             return -997;
@@ -1977,13 +1977,27 @@
```

`CreateBuffers` keeps a third power-of-two gate that 0004 did not reach, so 0005's adopt path offers a size the driver then rejects. With `fixed_buffer_size = false`, pinning the graph to 768 gives `Invalid buffersize 768 requested` and Live drops the device. reproduced 3×, and absent from the same run with `PIPEASIO_FIXED_BUFFERSIZE=on`, which is what the F1 verification recorded.

patches/pipeasio/0005-converge-on-a-foreign-quantum-predict-adopt-mute.patch
```diff
@@ -0,0 +1,439 @@
...
+         {
+             spa_hook_remove(&c->registry_listener);
+@@ -643,6 +654,11 @@
+     const size_t bsize_samples = c->buffer_size;
+     const size_t bsize_bytes   = bsize_samples * sizeof(audio_sample_t);
+ 
++    /* Escape hatch for the quantum-mismatch fail-safe (issue #49): "on"
++     * keeps audio flowing at mismatched sizes, the pre-1.2.2+0005 behaviour. */
++    const char *allow_mm     = getenv("PIPEASIO_ALLOW_QUANTUM_MISMATCH");
++    c->allow_quantum_mismatch = allow_mm && !strcmp(allow_mm, "on");
```
ClickSentinel:
```suggestion
+    c->allow_quantum_mismatch = allow_mm && !strcasecmp(allow_mm, "on");
```
`audio.c` parses this with `strcmp` while `asio.c` uses `lstrcmpiA`, which passes `NORM_IGNORECASE`. `PIPEASIO_ALLOW_QUANTUM_MISMATCH=ON` therefore disables adoption but not muting — permanent silence, the exact failure the hatch exists to prevent. Every other `PIPEASIO_*` variable uses `strcasecmp`. The added include is required: `audio.c` does not include `<strings.h>` and the unix side builds with `-Werror=implicit-function-declaration`, so the swap alone breaks the build at step 4/8.

patches/pipeasio/0005-converge-on-a-foreign-quantum-predict-adopt-mute.patch
```diff
@@ -0,0 +1,439 @@
+diff --git a/src/audio.c b/src/audio.c
+--- a/src/audio.c
++++ b/src/audio.c
+@@ -276,7 +276,10 @@
```
ClickSentinel, line 4:
```suggestion
@@ -31,5 +31,6 @@
 #include "wine/debug.h"
 
 #include <stdlib.h> /* getenv for PIPEASIO_DEBUG */
+#include <strings.h> /* strcasecmp */
 
 /* Raw stderr logging; TRACE is gated by PIPEASIO_DEBUG. */
@@ -276,7 +276,10 @@
```
`audio.c` parses this with `strcmp` while `asio.c` uses `lstrcmpiA`, which passes `NORM_IGNORECASE`. `PIPEASIO_ALLOW_QUANTUM_MISMATCH=ON` therefore disables adoption but not muting — permanent silence, the exact failure the hatch exists to prevent. Every other `PIPEASIO_*` variable uses `strcasecmp`. The added include is required: `audio.c` does not include `<strings.h>` and the unix side builds with `-Werror=implicit-function-declaration`, so the swap alone breaks the build at step 4/8.

PR #120, opened 2026-08-02
ableton-linkd: quiet logging + allowing installers to kill blocking ableton-wine processes pre-upgrade

shibco:
Two fixes in one because they were kinda related:

A user asked why `ableton-linkd` needed to be force quit, but kept restarting automatically. After reviewing this (and finding a separate issue of `ableton-linkd` being overly chatty). What I realised after a few clarifying messages with the reporter, was that they were trying to update ableton-linux and couldn't because it was blocked by still-running ableton-wine processes. So the user hunted down and killed the one visible "ableton" process, linkd, which the unit brought straight back.

This PR includes two things"
1. `ableton-linkd` now now longer prints the same messages over and over, and logging is now event-driven. You can restore the old behaviour with `--verbose` if you want that for some reason.
2. The installer now stops Live, the prefix wineserver, and any leftover runtime processes itself, asking for confirmation first when Live is open in a terminal session. It also stops `ableton-linkd` before swapping its binary and restarts it after, so updates actually take effect instead of the old process running from a deleted `inode` until reboot.

### DIFF
--- a/scripts/install.sh
+++ b/scripts/install.sh
@@ -105,25 +105,28 @@ for p in $(runtime_pids); do
     esac
 done
 if ableton_up; then
-    echo "== stopping running Ableton processes ($(runtime_pids | wc -l)) =="
+    echo "== stop processes using the installed runtime =="
+    echo "   $(runtime_pids | wc -l) found"
     # Closing Live discards unsaved work, so require an explicit yes.
     # -r and -w cannot ask that: they stat a 0666 device node and pass
     # even with no controlling terminal, and the printf would then fail
     # ENXIO and abort the install under set -e. Opening it is the only
-    # honest test. A timeout, an EOF or a missing terminal all mean
-    # nobody consented, so refuse. Leftover wineserver and winedevice.exe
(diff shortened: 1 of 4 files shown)


scripts/install.sh
```diff
@@ -55,10 +55,37 @@ else
...
+    pkill -f '[A]bleton Live.*\.exe|[P]ush2DisplayProcess.exe' 2>/dev/null || true
+    if [ -x "$OPT/$NAME/bin/wineserver" ]; then
+        WINEPREFIX="${ABLETON_WINEPREFIX:-$HOME/.wine-ableton}" \
+            "$OPT/$NAME/bin/wineserver" -k 2>/dev/null || true
+    fi
+    for _ in $(seq 1 20); do
+        pgrep -f "$OPT/$NAME" >/dev/null 2>&1 || break
+        sleep 0.5
+    done
+    pkill -9 -f "$OPT/$NAME" 2>/dev/null || true
```
ClickSentinel: Would really like to see this be a `winserver -k` with fallback to pkill if fails

PR #121, opened 2026-08-02
Multi audio device hardening

shibco:
This PR is part of an ambitious series of patches called _Performance Moonshot_ that aims to drastically improve the capability, speed and stability of ableton-linux.

It is therefore dependent on #118 to be merged before this one.
 
When Live's input and output are separate devices, each has its own sample clock. 

PipeWire lets one device set the timing and continuously adjusts the other, called the follower, to keep it in step. That adjustment is silent while the follower has spare buffer room to absorb timing fluctuations, and this is probably the source of some of the audio artefacting people have reported.

This patch updates the behaviour so that when a session plays across two devices, we raise the follower's spare room via `api.alsa.headroom`, and put the old value back when the session ends or the devices change. 

Additionally, the property can be changed or disabled entirely on a live node through its Props parameter `PIPEASIO_FOLLOWER_HEADROOM`.


ClickSentinel (commented):
Reviewed at 3d3f512: built (audit 92/92) and exercised at runtime through a minimal ASIO host against a real two-device graph, PipeWire 1.5.85 (stock Pop!\_OS 24.04). Driver level only — Live was not in the loop, so the latency reporting the crackle note already lists as open stays open here too.

## Blocking

- 0007 never raises on PipeWire 1.5.85. `audio_hr_tick` probes with `pw_node_enum_params(..., SPA_PARAM_Props, 0, 1, NULL)`; here Props enumeration yields two objects and `clock.name` / `api.alsa.headroom` sit only in the second (`pw-cli enum-params <node> Props` shows this). `clock_name` stays empty, every slot skipped, no diagnostic. Instrumented build: `running=1 two=1 gclock='api.alsa.c-0'`, both slots `probed=1 clock=''`, indefinitely. **Fix:** enumerate all Props objects (`max` > 1, or `pw_node_subscribe_params`) — the handler already accumulates across events. Verified with only `max=16`: raise in ~6 s on the correct follower, 512 confirmed, clean-exit restore works; also stops `cur_headroom` recording a stale 0 to restore. **_1.5.85 is current stock Pop!\_OS 24.04, so this is a mainstream deployment._**

## Should fix

- PR body: `PIPEASIO_FOLLOWER_HEADROOM` is an env var read once at init, not a Props parameter on a live node — that's `api.alsa.headroom`. Setting a Props param by that name does nothing (verified). **Fix:** reword the last paragraph — the env var picks the amount or disables; the driver applies it via the node's `api.alsa.headroom` Props param.
- build-audit fingerprints `pipeasio64.dll.so`, but Wine loads `pipeasio.dll.so` (the spec resolves pipeasio64 → builtin pipeasio; confirmed in `/proc/<pid>/maps`). Affects all seven pipeasio rows. **Fix:** point the FINGERPRINTS rows at `pipeasio.dll.so`, or check both files.

## Comment

- Only exact `off` / `0` disable; `'off '`, `OFF`, `no`, and `-1` (the internal disabled sentinel) clamp to 16 and enable. The WARN fires, but "tried to disable" becomes "enabled at minimum". **Fix:** trim + case-fold before the compare, and treat unparseable or negative values as off rather than clamping up.
- An unclean exit leaves the 512 raise on the shared device until the node is recreated (verified); other clients keep the latency. **Fix:** a line in the crackle note; inherent to the approach.
- In Dummy-Driver windows neither device matches the graph clock, both get raised, and each WARN claims the device "follows the other one's clock". **Fix:** when the match fails for both, name the actual graph clock in the WARN instead of "the other one".
- Registration is complete and verified: SERIES.sha256, fingerprints, note addenda; patch-check and the audit pass; the series applies to vendor/pipeasio-1.2.2 with `--fuzz=0`; the base had not moved under the PR. No action.

shibco:
This is a big single commit, sorry, it also covers a lot of docs.

ClickSentinel (approved):
All blocking items are resolved. Suggestions only:
 
- Negative and zero values disable silently. `abc` gets a WARN, `-512` gets nothing — a typo reads as "off". **Fix:** same WARN shape as the unparseable case.
- `any_sets_timing` is computed from whichever slots have replied. Both `enum_params` go out in the same tick, so in practice both land before the next one — every run here used the "follows the other one's clock" wording. If one reply is late, the first raise WARN would name the graph clock instead, and it is a one-shot message. Read, not observed. **Fix:** skip the raise until both slots are probed, or nothing.
- The docs pass leaves two nested-paren sentences where an em-dash pair became an outer paren: `AUDIO-LATENCY-ECOSYSTEM.md` "the priority ladder (PipeWire data threads (rtkit/rt.prio), …)" and `PROJECT-TECHNOLOGIES.md` "the open risks (Linux throttles … synchronous calls to (`notes/…:31-41`))". A few sentences also now carry two colons. **Fix:** recast those to commas or a sentence break.
- `patches/BASE.txt` never mentions `patches/pipeasio/`. Its "Applying the series" section describes only `patches/00*.patch` via `git am`; the pipeasio loop at `container-build.sh:138` is undocumented. Pre-existing and also true on main, so not this PR's regression, but this is the second PR to grow that directory and it is now seven patches. Numbering is not at risk — `SERIES.sha256` and the audit's `n_asio` both cover it. **Fix:** a paragraph in BASE.txt.
- The TROUBLESHOOTING ntsync rewrite reads much better, but "if the first command reports that the module does not exist, your kernel is too old" drops the case the old text named — a 6.14+ kernel built without `CONFIG_NTSYNC` — and "install your distribution's newest kernel" is a dead end when that is still older than 6.14. **Fix:** keep the "move to a kernel that provides the module" phrasing.

shibco:
Good suggestions, I will fix them and merge.

PR #123, opened 2026-08-02
Add benchmark projects with OSC for interactive puppeteering

shibco:
This PR adds additional sets to simulate a full collection of projects with various configurations of instruments, tracks and audio.

Each has our small OSC max-for-live device, allowing remote control of the projects.


shibco:
As an aside this will be eventually paired with an implementation that works with the Ableton Extensions SDK

PR #124, opened 2026-08-02
Defeating the Wine vs Live GPU Denylist problem, take 3 

shibco:
Patch 0061 was meant to cover table misses by describing a card from its driver's own name, but it stands down when the driver reports no video memory. 

But, of course, this didn't fix all possible circumstances, did it?

Turns out I was wrong because I thought that Live's denylist is based on the device name. It seems like it isn't, and that instead it's based off the graphics device's PCI ID

Wine's device table skips 24 Intel devices between Skylake and Amber Lake, because why wouldn't Wine do this? Amazing stuff. 

Among those skipped are entries like `UHD Graphics 630 0x3e92`, and other non-niche devices. When the device is not in the table, wined3d guesses a card identifier from its reported feature level, and for Intel that guess seems to be device `0x0162`, "`Intel(R) HD Graphics 4000`". Which we have already determined is a beautiful central star of Live's GPU denylist.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,7 +13,7 @@
 - Graphics cards missing from Wine's table keep their real identity even when
   the driver reports no video memory (Wine patch 0067). Wine reads that figure
   from an optional graphics-driver feature, and a driver that does not offer it
-  used to cost the card its identity, falling back to the invented 2012 one.
+  used to make Wine report the card under the invented 2012 identity instead.
 - Added `WINE_D3D_FORCE_GPU_RENDERING=1`, which offers Live's GPU renderer on
   the older Intel graphics Live refuses by model (Wine patch 0068). Run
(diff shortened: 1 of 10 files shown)


shibco:
Wait a minute, we're in Wine. We can do _anything_. Like, we can tell Ableton Live that this is any card we want.

Here's a flag that does just that:

`WINE_D3D_FORCE_GPU_RENDERING=1 ableton-live`

It tells gives Live a device identifier that its happy with,  so the GPU renderer becomes available on hardware Live otherwise refuses outright.

(it would be very cool to try this on machines with older cards that Ableton very specifically doesn't support, I wonder if there are certain GPUs that previously couldn't run Live fully, and now can...)

We also don't want to mess up Ableton's crash logs so in the interest of transparency, we mark the fact that the card is actually another card in a trenchcoat. For example, `Intel(R) HD Graphics 4000 (reporting as Intel(R) UHD Graphics 630)`.

PR #125, opened 2026-08-02
Patch 0069: prevent monitor-sized Live windows from losing resizability

trendwhore:
## Summary

Keep the selected Live window resizable when it remains captioned but covers its monitor. Borderless fullscreen and unselected windows retain the existing behavior.

## Cause

Wine clears `WINE_SWP_RESIZABLE` when the window geometry covers the monitor, even when Live retains `WS_CAPTION` and `WS_THICKFRAME`. X11 consequently advertises identical minimum and maximum sizes.

This was easy to reproduce under Niri because my configuration does not use titlebars or window borders, allowing a captioned window's geometry to match the monitor exactly.

## Validation

Tested under Niri and KWin/Xwayland. Captioned monitor-sized windows remain resizable and expose Live's application-defined minimum without a maximum. Fullscreen entry and exit continue to work correctly.

`WINE_WIN32_RESIZABLE_CLASS=off` reproduces the fixed-size behavior for comparison.

### DIFF
--- a/BUILDING.md
+++ b/BUILDING.md
@@ -73,6 +73,8 @@ make verify
 - `ABLETON_DCOMP=off` disables DirectComposition for one launch.
 - `WINE_X11_FORCE_OFFSCREEN_CLASS=off` disables the default Max for Live
   selection-flicker fix for one launch.
+- `WINE_WIN32_FULLSCREEN_CLASS=off` disables the default Live fullscreen
+  layout and exit-state fix for one launch.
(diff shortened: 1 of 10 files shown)


trendwhore:
Tested on Sway, MangoWM, Gnome. Works as expected :D

shibco:
Taking a look now!

shibco:
Passed the audit! welcome aboard!

PR #114, opened 2026-08-01
Patch 0065: fix Live fullscreen layout and exit

trendwhore:
## Summary

- Add Wine patch 0065 to normalize monitor-covering geometry for the launcher-selected Ableton Live window class.
- Exclude Live's still-attached native menu from fullscreen layout and painting, keeping rendered content and pointer coordinates aligned.
- Retire an already-satisfied X11 configure request during fullscreen exit so `_NET_WM_STATE_FULLSCREEN` is removed before restoring windowed geometry.
- `WINE_WIN32_FULLSCREEN_CLASS=off` is available for A/B comparison.

## Cause

Live retains its native Win32 menu while entering fullscreen. Wine consequently reserves a menu band even though Live expects the client area to cover the output, producing shifted or clipped content and offset pointer input.

On fullscreen exit, Wine can also retain a pending configure request whose rectangle has already been applied. It then waits for another configure before removing `_NET_WM_STATE_FULLSCREEN`, while the window manager waits for fullscreen removal before granting the restored rectangle. Moving or resizing the window breaks that wait accidentally.

Patch 0065 scopes both corrections to the exact top-level class selected by the launcher. VST windows continue to open, render, and resize normally.

## Validation

- The complete Wine 11.13 patch stack built successfully, and the build audit passed all 88 checks, including the new win32u and winex11 fingerprints.
- `View -> Full Screen` and F11 entry and exit were repeatedly verified on both Niri with xwayland-satellite and Plasma/KWin with Xwayland.
- Fullscreen hides the title bar, native menu, and desktop panel while keeping client content and pointer input aligned.
- Exiting fullscreen immediately restores normal floating geometry without stale fullscreen visuals; the window can be moved and resized without an extra drag or configure event.

Fixes #42

PR #127, opened 2026-08-03
PR124: Review/pr124 followups

ClickSentinel:
Three follow-ups from reviewing #124, based on 466e5fa so they land in the branch. No rebase — #114's 0065 conflicts with `BASE.txt`, `SERIES.sha256` and `build-audit.sh`, easier for you to do it once. Each commit carries its own `SERIES.sha256` line and applies independently. Audit passes 91/91.

**`5aa607b`** — `substitute_gpu_device_id` ran at the tail of `query_gpu_description`, so it was skipped when that returned NULL. On that path `wined3d_adapter_init_gl_caps` settles the description from `wined3d_guess_card` — for Intel, the fabricated `0x0162` the substitution exists to replace. The flag was a silent no-op there. Moved to where the description is final. Forced the branch with `LIBGL_ALWAYS_SOFTWARE=1`: no trace at all before, `Not substituting a device ID for vendor 10de` after.

**`e393024`** — `forced_description` was 192 bytes but `DXGI_ADAPTER_DESC1` truncates at 128, so a card name over 86 chars would drop the disclosure while the substituted ID still went through. Buffer sized to DXGI's field, card name bounded. No change today (63 chars) — reasoned, not exercised, since I can't produce a long enough name here.

**`ec60941`** — "`GL_NVX_gpu_memory_info` … which Mesa does not" is wrong as a universal; Mesa implements it per driver. radeonsi on Mesa 26.1.6 reported 20 GB over EGL here, so 0061's guard is no obstacle there. Restated as conditional across both patch headers, `BASE.txt`, notes and changelog. AMD only — no Intel iGPU here, so iris is untested. Neither patch depends on the answer.

Testing:

Real Live 12.4.3 on this build, isolated prefix clone, Intel identity via `HKCU\Software\Wine\Direct3D`.

Flag off (`8086:0162`) — `Can't use GPU renderer: Intel(R) HD Graphics 4000: Unexplained slow UI at zoom-level 100% and/or crashes`, setting greyed out with that reason.

Flag on — `TD3dSurface: Adapter: 'Intel(R) HD Graphics 4000 (reporting as Intel(R) UHD Graphics 630)' (8086:3e9b rev0)`, setting available and on.

The name was byte-identical in both runs; only the device ID differed, and the outcome flipped. Your disassembly finding, confirmed black-box.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,8 +12,8 @@
   the real device.
 - Graphics cards missing from Wine's table keep their real identity even when
   the driver reports no video memory (Wine patch 0067). Wine reads that figure
-  from a source only NVIDIA's driver provides, so on Intel and AMD it was
-  always absent and the card fell back to the invented 2012 identity.
+  from an optional graphics-driver feature, and a driver that does not offer it
+  used to cost the card its identity, falling back to the invented 2012 one.
 - Added `WINE_D3D_FORCE_GPU_RENDERING=1`, which offers Live's GPU renderer on
   the older Intel graphics Live refuses by model (Wine patch 0068). Run
   `WINE_D3D_FORCE_GPU_RENDERING=1 ableton-live` to try it. Diagnostics sent to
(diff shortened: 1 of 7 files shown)


shibco:
This is good to merge!

PR #129, opened 2026-08-03
Make ableton-linux's various system services better citizens

shibco:
This PR is based off feedback from @jackson-57. It ensures that the new optional RT scheduler and `ableton-linkd` only run when Ableton Live is actually open, and are offline at all other times.


shibco:
I am not good at diffs.

scripts/setup-link.sh
ClickSentinel, line 1: This records version 4 even when the migration above never ran — it sits behind three earlier conditions (daemon installed, unit file found, `daemon-reload` succeeds), and `prior_version -lt 4` at line 105 then reads every later run as a deliberate opt-in. Those hosts keep starting the daemon from login while the script prints "the launchers start ableton-linkd with Live".

Reachable without odd host state: `--no-link` writes `declined\n4` in `setup-run-header.sh` without running this script, so the later `--link` skips the migration. Reproduced under a scratch `HOME` — with the marker forced to `3` the same run prints `earlier setups ran ableton-linkd from login; disabled that`. `daemon-reload` failing over SSH is the same shape.

The hook removal two steps up already sets the convention: "Do not mark this version configured with the hook still in place." Same reasoning here - write the version only when the anchor step completed.

scripts/ableton-live
```diff
@@ -792,11 +793,25 @@ if [ "${ABLETON_RT:-on}" != off ] && command -v chrt >/dev/null 2>&1 \
...
+# runs, through power-profiles-daemon (tuned-ppd answers the same CLI). The
+# hold needs no root, shows up in the desktop's power menu, and is released
+# when Live exits or crashes. This replaces the boot-time governor unit that
+# setup-realtime.sh installed before 2026-08; ABLETON_POWER=off skips it for
+# one launch. The probe runs the exact call with `true`, so a missing daemon,
+# a host without a performance profile, or an old powerprofilesctl all fall
+# back to launching without the hold.
+if [ "${ABLETON_POWER:-on}" != off ] && command -v powerprofilesctl >/dev/null 2>&1 \
+    && powerprofilesctl launch -p performance -r "Ableton Live session" -i ableton-live -- true >/dev/null 2>&1; then
+    run=(powerprofilesctl launch -p performance -r "Ableton Live session" -i ableton-live -- "${run[@]}")
```
ClickSentinel, line 814: The hold is released immediately on the `wine start` paths (lines 860 and 862). `powerprofilesctl`'s `_launch` holds only across `Popen(...).wait()` on its direct child, and `start.exe` returns as soon as it hands off — `WaitForSingleObject` in `programs/start/start.c` is gated on `SEE_MASK_NOCLOSEPROCESS`, which only `/w`, `/wait` and `/exec` set. So `ableton://` URLs and `.auz` files, which the desktop entries route here, get a full session with no hold.

`chrt` survives those paths because `spawn_process` double-forks and SCHED_RR is inherited; the D-Bus hold can't follow an orphaned grandchild. `max9` is fine — it execs the Max exe directly.

Live already being up is fine — the guard at line 132 lets `wine start` forward to the running instance, whose original launcher still holds the profile. It's the cold start through a URL or `.auz` that gets no hold.

scripts/setup-realtime.sh
```diff
@@ -87,37 +90,34 @@ if [ -z "$DESTDIR" ]; then
...
+# Earlier versions set the performance governor here and enabled a unit that
+# reapplied it at every boot: a permanent battery cost for hosts that run
+# Live occasionally. The launcher now holds the "performance" power profile
+# through power-profiles-daemon only while Live runs (ABLETON_POWER=off skips
+# the hold for one launch). This step only removes what earlier versions
+# installed.
+if [ -n "$DESTDIR" ]; then
+    echo "   staged mode — nothing to stage (the launcher holds the profile per session)"
+else
+    if [ -f "$OLD_GOV_UNIT" ]; then
```
ClickSentinel, line 103: This removal only reaches people with a checkout — `setup-realtime.sh` isn't in `make-installer.sh`'s staged list, and `uninstall.sh` never mentions the unit. `uninstall.sh` already has the section for root-owned leftovers (the NM hook, the firewall port); the `systemctl disable --now` / `rm` pair belongs there too, and TROUBLESHOOTING.md should give the commands rather than only "run setup-realtime.sh again".

notes/ABLETON-WINE-LINK.md
```diff
@@ -58,31 +69,35 @@ SDK, and installs it at:
...
+exits on its own. Live counts as a peer while Link is enabled in its
+preferences, so the daemon stays up through a Live restart or crash and
+stops soon after you finish working. `--linger` changes the wait.
 
 Its modes are:
 
-- No arguments: run in the foreground and write to stderr when the peer
-  count, tempo, or transport state changes. The systemd unit uses this mode.
-  A quiet session writes nothing.
+- No arguments: run in the foreground and write status to stderr every ten
```
ClickSentinel: This describes `--verbose`, not the default: the daemon still logs on change and gates the ten-second status behind `verbose`. The `--verbose` bullet was replaced by `--linger`, but the flag is still in the parser and the usage text — looks like the doc half of "restore --verbose after the merge from main".

scripts/ableton-live
```diff
@@ -792,11 +793,25 @@ if [ "${ABLETON_RT:-on}" != off ] && command -v chrt >/dev/null 2>&1 \
...
+# CPU speed for the session: hold the "performance" power profile while Live
+# runs, through power-profiles-daemon (tuned-ppd answers the same CLI). The
+# hold needs no root, shows up in the desktop's power menu, and is released
+# when Live exits or crashes. This replaces the boot-time governor unit that
+# setup-realtime.sh installed before 2026-08; ABLETON_POWER=off skips it for
+# one launch. The probe runs the exact call with `true`, so a missing daemon,
+# a host without a performance profile, or an old powerprofilesctl all fall
+# back to launching without the hold.
+if [ "${ABLETON_POWER:-on}" != off ] && command -v powerprofilesctl >/dev/null 2>&1 \
+    && powerprofilesctl launch -p performance -r "Ableton Live session" -i ableton-live -- true >/dev/null 2>&1; then
```
ClickSentinel: Right shape for the probe — exact flags, so an old `powerprofilesctl` without `-i` degrades cleanly. Worth a `timeout` around it though: these are sync D-Bus calls, so a wedged daemon stalls each launch for GLib's 25 s default, twice over.

ClickSentinel (changes requested):
One blocking issue (the version marker can strand the migration, and nothing re-examines it afterwards — only a bump to version 5 would), two behavioural gaps, and some smaller notes inline.

notes/ABLETON-WINE-LINK.md
ClickSentinel: Ran the built binary: with no arguments it logs the enable line and one status, then nothing for 25 s; with `--verbose` it logs at 10 s and 20 s.

tools/ableton-linkd.cpp
```diff
@@ -291,9 +320,23 @@ int main(int argc, char** argv)
...
+    int linger_secs = kDefaultLingerSecs;
     const char* verbose_env = std::getenv("ABLETON_LINKD_VERBOSE");
     bool verbose = verbose_env && *verbose_env && std::strcmp(verbose_env, "0") != 0;
 
+    /* Environment first, flags override: the launchers start the daemon
+     * without arguments, so the environment is the per-launch override. */
+    if (const char* env = std::getenv("ABLETON_LINKD_LINGER")) {
+        double secs;
+        if (parse_double(env, 0.0, kMaxLingerSecs, secs)) {
+            linger_secs = static_cast<int>(secs);
```
ClickSentinel: `--linger 0.5` parses, then `static_cast<int>` turns it into `0` — the never-exit sentinel. Verified: the daemon ran indefinitely and also skipped the `session scope: exiting after N s` startup line, since that's gated on `linger_secs > 0`, so nothing signals that the linger was dropped. Same at line 380, and `ABLETON_LINKD_LINGER=0.5` takes the same path. Rejecting non-integers, or rounding up to 1, would both do.

scripts/setup-link.sh
```diff
@@ -87,24 +98,34 @@ else
...
-    echo "OK: Link setup complete; the Live launcher will start ableton-linkd"
+if [ "$anchor" = session ]; then
+    echo "OK: Link setup complete; the launchers start ableton-linkd with Live"
+    echo "    (always-on anchor instead: systemctl --user enable --now ableton-linkd.service)"
 else
     echo "OK: Link setup complete; ableton-linkd anchor skipped"
 fi
 mkdir -p "$HOME/.local/share/ableton-wine"
-printf 'configured\n%s\n' "$LINK_SETUP_VERSION" > "$HOME/.local/share/ableton-wine/link-configured"
+printf 'configured\n%s\n' "$LINK_SETUP_VERSION" > "$marker"
```
ClickSentinel: Reproduced on a real systemd user session, not stubs. With the unit enabled and the marker at `configured 3` the script prints `earlier setups ran ableton-linkd from login; disabled that` and the unit ends up disabled. With the same unit enabled and the marker at `declined\n4` — which is what an update run with `--no-link` writes, before any `--link` opt-in — `[3/3]` prints nothing and the unit stays enabled. Both runs close with `the launchers start ableton-linkd with Live`, so the stranded host is told the opposite of its actual state.

scripts/setup-realtime.sh
```diff
@@ -87,37 +90,34 @@ if [ -z "$DESTDIR" ]; then
...
-if [ "$write_unit" -eq 1 ]; then
-    install_dropin "$GOV_UNIT" <<'EOF'
-[Unit]
-Description=ableton-linux: performance CPU governor
-[Service]
-Type=oneshot
-ExecStart=/bin/sh -c 'echo performance | tee /sys/devices/system/cpu/cpu*/cpufreq/scaling_governor'
-[Install]
-WantedBy=multi-user.target
+    if command -v powerprofilesctl >/dev/null 2>&1; then
```
ClickSentinel, line 113: Worth naming the cost on Pop!_OS and other System76 installs: `apt install power-profiles-daemon` removes `system76-power` and `system76-driver`, and leaves the rest of the System76 stack flagged for `autoremove`. Following this advice there trades the vendor's power management for the profile hold. Might be fine.

shibco:
Thanks, @ClickSentinel, this should cover your feedback.

ClickSentinel (changes requested):
Re-reviewed at `0f938e2`; all seven findings from `1853931` are addressed. Checked by running the code: `configure_link` extracted verbatim across five marker states, `start` vs `start /w` timed against a 12-second stand-in exe, the rebuilt daemon's `--linger` parser, and the declared-base merge. The D-Bus hold itself is unexercised — polkit denies `hold-profile` to a seatless shell — so it is argued from the start.exe wait it rides on.

One blocking regression inside the fix for the previous blocking finding.

## Blocking

**`--no-link` aborts the installer when no marker exists yet** — `setup-run-header.sh:211`. `baseline="$(sed -n 2p "$marker" 2>/dev/null)"` lost the `|| true`; the file runs under `set -euo pipefail` and `sed` exits 2 on a missing file, so the installer dies right after printing `-- Ableton Link setup skipped (--no-link)`, with no error of its own. `configure_link` sits at line 305, after `install.sh` and before `setup-prefix.sh`, so the runtime lands and the prefix is never created. Reachable on a first install with `--no-link`, and on an update of any host that never configured or declined Link.

Verbatim function against a scratch `HOME`:

| marker | result |
|---|---|
| absent (with or without the parent dir) | **exit 2** |
| `configured 5` / `configured 3` / `declined 3` | exit 0, baseline preserved |

So the baseline-preservation intent is right and only the absent case breaks. Fix: `|| true`, the form `setup-link.sh:89` already uses for `prior_version`.

## Should fix

**`notes/` still describes the version-4 migration.** `ABLETON-WINE-LINK-FIRSTCLASS.md:128` states the opposite of the code — "an enablement made after the version 4 migration is treated as that opt-in and left alone" — where `setup-link.sh:119` tests `prior_version -lt 5` and its comment says a marker reading 4 "cannot be trusted as a deliberate choice". Same for `FIRSTCLASS.md:119`, `:125`, `:172` and `ABLETON-WINE-LINK.md:39`. Fix: move all five to version 5, with `:125` reading versions 1 to 4.

## Comment only

**Headless installs never record a marker.** The new `exit 1` on a failed `daemon-reload` is the right call for not stranding the migration, but a seatless run then repeats `[1/3]` every update, including the `sudo ufw allow 20808/udp` prompt, and always closes on "!! Ableton Link was not configured". Reasoned, not run. Fix: let the firewall half record its own completion so only the anchor step retries.

## Holds

- The v5 marker gate is sound on every path: a missing daemon, a missing unit file and a failed `daemon-reload` all leave it unwritten. The `! command -v systemctl` branch recording normally is right — there is no unit state to check there.
- `--linger` is exact. `parse_secs` adds `v != floor(v)` over an already trailing-byte-exact `parse_double`, so `0.5` gives exit 2 on the flag and a warning plus the 900 default on the environment variable; `-1`, `off`, `604801`, `1.0000001` reject. `1e3` and `0x10` land on 1000 and 16, harmlessly.
- `start /w` covers both start.exe paths (`ableton-live:868`, `:870`) and nothing else needs it: `max9` execs the Max exe directly, and those two are the only holders of the profile. Timed — plain `start` returned in 0 s with the child alive, `start /w` after its full 12 s.
- The governor notice matches the file's existing handling of root-owned leftovers, which print rather than remove; the unit path agrees across `setup-realtime.sh:44`, `uninstall.sh:62` and `TROUBLESHOOTING.md:168`.
- `git merge --no-commit --no-ff upstream/main` is clean against the declared base; `af6f831` (#147) landed after this branch's last merge and does not collide. `bash -n` passes on all six changed scripts.

shibco:
Ah

shibco:
there. happy now? >: (

shibco:
clowned myself and wrote some bullshit that wasn't right after being an asshole about docs consistency.
got caught, fixed the docs

Issue #111, opened 2026-07-31
Installer hangs until Ableton USB Driver has exited


HurleybirdJr:
## What happened

I noticed this early on with the project when it was possible to run the `install-ableton-latest.run` script to update without reinstalling.

The script would update the wine instance, etc... then get stuck on this step:

```
== [1/5] initialise prefix at /home/hurleybirdjr/.wine-ableton ==
```

until I closed the Ableton USB Driver application via my taskbar, and then it could carry on no problem shortly after that.

I've double checked this is consistent every time, and each time I retry the installer via the `--update` flag, it always hangs on this step, that is until I close the Ableton USB Driver program.

## Steps to reproduce

1. Run the installer script `sh install-ableton-latest.run --update`
2. Let the installer do it's thing...
3. Get to step `[1/5]` and it will hang.
4. Wait however long you want, but until the moment Ableton USB Driver program closes, it stays on this step

## System summary
```
[SYSTEM]
PRETTY_NAME="CachyOS"
Linux 7.1.5-1-cachyos x86_64
CPU(s):                                  32
Model name:                              AMD Ryzen 9 9950X 16-Core Processor
               total        used        free      shared  buff/cache   available
Mem:            60Gi        27Gi        11Gi       345Mi        22Gi        32Gi
sys_vendor=ASUS
product_name=System Product Name

[GRAPHICS]
OpenGL vendor string: NVIDIA Corporation
OpenGL renderer string: NVIDIA GeForce RTX 3080 Ti/PCIe/SSE2
OpenGL core profile version string: 4.6.0 NVIDIA 580.173.02
OpenGL version string: 4.6.0 NVIDIA 580.173.02
desktop=KDE
session=wayland

[AUDIO]
pipewire=active
pipewire-pulse=active
wireplumber=active
pipewire
Compiled with libpipewire 1.6.8
Linked with libpipewire 1.6.8
wireplumber
Compiled with libwireplumber 0.5.15
Linked with libwireplumber 0.5.15
Found "settings" metadata 31
update: id:0 key:'log.level' value:'2' type:''
update: id:0 key:'clock.rate' value:'48000' type:''
update: id:0 key:'clock.allowed-rates' value:'[ 48000 ]' type:''
update: id:0 key:'clock.quantum' value:'1024' type:''
update: id:0 key:'clock.min-quantum' value:'32' type:''
update: id:0 key:'clock.max-quantum' value:'2048' type:''
[...]

PR #130, opened 2026-08-03
PR119: installer family

ClickSentinel:
Live 12 ships an Inno Setup installer; Live 11 ships a WiX Burn bundle.

Burn does take `/SILENT` (its own synonym for `/quiet`) and `/NORESTART`, but it has no `/SUPPRESSMSGBOXES` and no `/MERGETASKS`. So on Live 11 the current flags mean the USB audio driver installs and fails regardless — #111 is untouched there — and `/SILENT` selects Burn's quiet mode, which shows no window at all rather than the progress bar the message promises.

Identify the family and send the flags only where they mean something. Live 12 keeps your invocation exactly; Live 11 falls back to no flags, byte-identical to main, so the behaviour change stays scoped to Live 12.

Detection is ordered so the expensive case can't happen. `.wixburn` is a PE section ~630 bytes in, so 4k rules Burn out however large the installer is — and Burn is the case with no Inno marker to stop at. The Inno marker is ~680 KB in, so 4M covers it. Both reads use process substitution rather than a pipe: `grep -q` exits at the first match, `head` then takes SIGPIPE, and under `pipefail` the pipeline would report 141 and pick the wrong branch.

Verified end to end through a built kit with an isolated `HOME`:

| | result |
| --- | --- |
| Live 12.4.2 | progress window, no clicks, driver skipped — no `tlsetupfx.log`, no `AbletonPush` registry entries — exit 0 |
| Live 11.3.25 | wizard as before, `Applied non-vital package: Ableton_Push_Audio_Driver.msi, error: 0x80070643. Continuing...`, exit 0 |
| unrelated binary | falls through to no flags |

### DIFF
--- a/scripts/setup-prefix.sh
+++ b/scripts/setup-prefix.sh
@@ -648,12 +648,14 @@ cat <<EOF
 Remaining steps (you supply Ableton + your own license):
 
   1. Install Live (any edition) through THIS wine (plain wine reads
-     WINEPREFIX, not the ABLETON_* launcher variables). The flags let the
-     installer run by itself and skip Ableton's Windows USB audio driver,
-     which does nothing on Linux:
+     WINEPREFIX, not the ABLETON_* launcher variables). For Live 12 the flags
+     let the installer run by itself and skip Ableton's Windows USB audio
+     driver, which does nothing on Linux:
        WINEPREFIX=$WINEPREFIX \\
-       $WINE_ROOT/bin/wine "/path/to/Ableton Live NN Edition Installer.exe" \\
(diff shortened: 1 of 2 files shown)


shibco (changes requested):
Two small fixes, comments inline.

ClickSentinel:
> Two small fixes, comments inline.

Fixed the 2 bugs. burn header is found early in the file, always >4kb in. Regardless of .exe size, it will not crawl the whole file. And correct about burn flag availability. 

Only other recommendation I would make is breaking out a new `install.sh` for this procedure. Right now the user has to run wine commands with flags, but we already have the variables and the tools to write a script that will do the right thing regardless if ableton 11 or 12 .exe is provided.

Issue #122, opened 2026-08-02
Max 4 Live devices interfere with Live's interface


shibco:
When adding a m4l device to a track, the behaviour of dragging sliders with your mouse (eg the volume slider in session view) becomes deeply erratic.

I've just witnessed this directly and am keeping this open to investigate.

This was seen on a Fedora environment with an additional high CPU DSP usage. (25%)

trendwhore:
I can reproduce this. After loading a Max for Live device, Live’s sliders begin dragging erratically. Merely hovering over the device does not help, but letting its embedded JUCE window receive focus once immediately restores normal slider behaviour.

I isolated the focus change from the mouse click by programmatically transferring focus to the JUCE child without sending any mouse-button or capture events:

```text
WM_KILLFOCUS
  window: Live main window
  new focus: JUCE child
  mouse capture: none

WM_SETFOCUS
  window: JUCE child
  previous focus: Live main window
  mouse capture: none
```

The next manual fader drag was smooth. Focus returned to Live’s main window during that drag, but the problem did not return.

I also traced `SetFocus` while loading the device from a fresh session:

```text
SetFocus(Live main window)
  caller: Ableton Live executable

SetFocus(JUCE child)
  not observed
```

This narrows the starting point for a fix to the focus/activation sequence when the embedded JUCE child is created. The next step is to determine whether a narrow Wine compatibility fix can trigger the missing initialization without changing normal focus behaviour.

trendwhore:
https://github.com/user-attachments/assets/9d61ae2a-a9d5-4e83-ad95-69e622940aa0
example of focus change and slider behaviour

shibco:
Fixed!

PR #132, opened 2026-08-03
Fix instances where Max4Live devices interfere with Live's UI

shibco:
This one turned out to be kinda easy.

Loading a Max for Live device embeds a device UI window owned by a second process into Live's window tree. 

If the window manager briefly focuses that window during the embed, the host process is handed Wine's cursor-clip pointer grab, and the release for that pointer grab never reaches it. Instead, the notification goes only to the new foreground thread and the desktop thread, while the host's own `FocusOut` is dismissed by the re-parenting and `WM_STATE` guards. 

As a result, the Host window thinks that it still clips the cursor, and from then on it converts every physical mouse delta into synthesized pointer input over Live's cursor stream.

This PR forces `winex11` to release its local clipping state at every point where a process can observe that the `X` input focus left it, and logs "lost X focus to another client while clipping" when that actually releases.

### DIFF
--- a/notes/ABLETON-WINE-M4L-INPUT-INJECTION.md
+++ b/notes/ABLETON-WINE-M4L-INPUT-INJECTION.md
@@ -1,0 +1,170 @@
+# M4L devices make Live's sliders erratic (issue 122)
+
+This note explains why loading a Max for Live device can corrupt mouse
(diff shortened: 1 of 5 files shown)


shibco:
Fixes #122

PR #137, opened 2026-08-04
feat(scripts): add Live auto-downloader script

yioannides:
I created a small shell script for retrieving and downloading the latest available package for Ableton Live.

The method is based on curl'ing links for Ableton's package repo (with links previously and publicly available on the Ableton subreddit) by searching for existing Live versions in reverse until the script doesn't return an error.

The link structure is somewhat similar for Suite, Standard, Lite and Intro, so it supports these, but I wasn't able to find how beta links are stored (probably privately available through their beta program).

The goal is for the user to be able to use the `--suite` / `--standard` / `--lite` / `--intro` flag (instead of providing their own download `.zip` file of the Ableton Live installer), auto-download Live and continue with the installation as usual.

### DIFF
--- a/scripts/live-downloader.sh
+++ b/scripts/live-downloader.sh
@@ -1,0 +1,40 @@
+#!/usr/bin/env bash
+
+# currently supported flags: --intro, --standard, --suite
+# couldn't find any exposed links for beta packages
+
+TYPE=${1#--}
+
+get_download_url() {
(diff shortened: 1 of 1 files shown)


shibco:
Thank you! This would be mighty convenient, but is on hold until I can confirm Ableton is ok with this.

PR #138, opened 2026-08-04
docs: state the mechanism instead of a figure of speech (0066, 0067)

ClickSentinel:
good text > bad text


shibco:
bad text dismissed, cya in hell!!!!!

PR #141, opened 2026-08-05
Add a defensive diagnosis for patch 0058 + hardening

shibco:
#100 and others describe a fault that pins Live on the GDI path at 98.9% of a CPU core against 25.2%, with a corrected user interface.

This patch is a revision for `0058` that further defends against this behaviour. At the same time, this PR adds defensive local telemetry that will help us understand any instance where the behaviour gets out of control.

- Wine now counts, for each Live window, how many frames it shows on the fast path and how many on the slow fallback path.
- Wine can tell a window resize from a real fault. A resize changes the compared sizes on every frame. A fault repeats the same two sizes on every frame.
- When the same two sizes repeat for about two seconds, Wine writes a warning. A second rule catches faults that alternate: more than half of all frames on the fallback path for ten seconds.
- The warning prints once. It asks the user to open an issue and to copy one log line. That line holds the data we need: both sizes, the display scale state, the window type, the graphics card, the session type, and the desktop.
- When Live closes a window that ever used the fallback path, Wine writes one summary line with the totals.
- The launchers keep this one warning channel open. All other Wine debug output stays off, as before.
- `TROUBLESHOOTING.md` has a new entry for CPU spikes during mouse input. The mechanism and the test plan are in `notes/ABLETON-WINE-GPU-RENDERER.md.`


TROUBLESHOOTING.md
```diff
@@ -61,6 +61,33 @@ If the setting is still greyed out on 2026.08.01.1 or newer,
...
+First, we need to check if the CPU spikes are specific to Live.
+When you notice a spike, open your terminal and run this command:
+
+- If you started Live from the desktop menu, run
+  `journalctl --user --since "-15min" | grep -i "present-size mismatch"`.
+
+If that command returns any text related to Wine or Ableton, please
+[open an issue](https://github.com/shibco/ableton-linux/issues) and
+paste the full log line that contains `Sustained present-size
+mismatch`.
```
ClickSentinel: The destroy summary added in this PR also contains `present-size mismatch`, and it
fires for transient fallbacks too (measured: a healthy session logged
`2 of 2334 presents fell back` and still produced the line at exit) — so this grep
as written sends healthy users to the issue tracker looking for a "Sustained" line
their log doesn't have. The trailing colon matches the evidence line only; verified
against real session logs: benign session 0 hits, faulted session exactly 1.

```suggestion
- If you started Live from the desktop menu, run
  `journalctl --user --since "-15min" | grep -i "sustained present-size mismatch:"`.

If that command prints anything, please
[open an issue](https://github.com/shibco/ableton-linux/issues) and
paste the whole line. It starts with `err:winediag:` and holds the
data we need.
```

ClickSentinel (approved):
Tested at 14f6727 on COSMIC/Wayland: build and audit clean, and the warning path exercised for real. With 0059's swapchain.c bracket reverted to resurrect issue 100's fault at 125%, the warning fires at exactly `identical-pair 120` and the evidence line alone pins the mechanism — `dst (0,0)-(1706,896) client (0,0)-(1365,717)`, both DPI contexts shown. On the healthy build, a machine-paced 26 s edge drag measured 613 of 3028 presents falling back (20%, longest run 7) with no warning: the resize/fault split holds with wide margin.

Three corrections from the runtime data:

1. Check 1 in the notes (revert the texture.c hunk) won't fire the warning — that hunk brings back the visual bar while the gate keeps agreeing. Reverting the swapchain.c bracket is what reproduces the sustained fallback; that's the rig above.
2. Check 3 can't pass as written: the destroy summary fires for transient fallbacks too — a near-idle session logged `2 of 2334 presents fell back` and still printed it. Either gate the summary on `warned` or amend the check; the TROUBLESHOOTING grep currently matches this benign line, see the inline suggestion.
3. The evidence line prints `session unknown desktop unknown` on every run. XDG_SESSION_TYPE and XDG_CURRENT_DESKTOP are present in the Unix environment but don't survive the env conversion, so `GetEnvironmentVariableA` never sees them. Without a Unix-side read the compositor field — the main triage axis for this bug class — is dead for all reporters.

Minor: `beta/tester-kit/run-session` (hard) and `scripts/max9` (default) still set plain `WINEDEBUG=-all`, dropping the evidence line for the people most likely to report. And if #153 merges first, `SERIES_GAPS[0070]` needs a manual fix at this branch's rebase — the audit never rechecks gap entries for numbers that exist.

Those items are your call, otherwise, approved.

shibco:
Ready for another look @ClickSentinel

ClickSentinel (commented):
Both must-fixes from the last pass are genuinely fixed, and the merge against `upstream/main` is clean.

**Should fix**

- **The idle reset disables the ratio rule for any fault that presents in bursts.** Modelling both revisions: an oscillating fault at 100% fallback in 8 s bursts separated by 6 s mouse pauses warns at t=14 s under `14f6727` and stays silent for the whole 80 s / 4801-of-4801-frame run under `0f57214`. The cliff is exactly `PRESENT_MISMATCH_IDLE_MS` — the same fault with 0.5 s pauses still warns at 10 s. The stable-run rule is untouched, so issue 100's persistent shape still fires in 2 s; what narrows is the backstop for the oscillating topology the patch message names as "the danger left". Fix: on an idle gap discard the partial window but keep `window_strikes` unless the gap is long. Tested at a 30 s strike-reset threshold — the two-bursts-across-a-60-s-gap false positive stays silent and the 8 s/6 s fault warns at 19 s.

- **The evidence line's trailing counters read 0 when the ratio rule fires.** `swapchain_present_mismatch_warn` prints the live `consecutive` and `stable_run`, but the ratio path runs on every present including agreeing ones, and the agrees branch zeroes both — an oscillating-fault report reads `consecutive 0 identical-pair 0`. This commit fixed the same problem for the rectangles by passing `last_dst`/`last_client` and left the counters beside them live. Fix: print `consecutive_max` / `stable_run_max`, already tracked and meaningful on both paths.

- **The recorded verification predates the code in this commit.** The notes now read "Runtime verification is done", but the destroy predicate changed from `mismatches` to `warned` here — check 3 ("no destroy summary must appear") is asserted against code that no longer exists, and the run behind it is the one that produced the transient summary. Check 2's cited evidence (613 of 3028 presents, longest run 7) came from a 96 DPI session while check 2 specifies 125%, where the drag false-positive risk is higher. Fix: re-run checks 2 and 3 against `0f57214` and state the scale each measurement ran at.

- **The tester kit lost its forced `WINEDEBUG`.** `beta/tester-kit/run-session` went from `export WINEDEBUG=-all` to `${WINEDEBUG:--all,+winediag}`, and the `unset` list above it does not include `WINEDEBUG` — a tester with an ambient value now silently changes the session the kit exists to make reproducible. Fix: add `WINEDEBUG` to that `unset` line, or keep the hard assignment here.

**Comment only**

- **The `journalctl` instruction is unverified for a desktop-menu launch.** `scripts/ableton-live` does not redirect its own stderr, so where the line lands is the desktop launcher's choice. On COSMIC, app scopes show stderr at `/dev/null` for some apps and an inherited pipe for others, and `systemd-run --user --scope` inherits stdio rather than routing to the journal. A user whose launcher chose `/dev/null` gets nothing from either instruction. Fix: tee the launcher's stderr to a fixed path and point TROUBLESHOOTING there — one instruction that survives every launch path.

**Held**

The XDG fix is correct and measured: `is_special_env_var` in `dlls/ntdll/unix/env.c:331` prefixes every `XDG_` with `WINE_HOST_`, and a probe under the built runtime read `WINE_HOST_XDG_SESSION_TYPE="wayland"` and `WINE_HOST_XDG_CURRENT_DESKTOP="COSMIC"` with both bare names absent — the `len >= sizeof(buf)` guard is right against both of `GetEnvironmentVariableA`'s return conventions. `+winediag` is not a spam channel: the base tree has 58 winediag call sites, 55 `ERR_` / 2 `WARN_` / 1 `FIXME_`, no `TRACE_`. All five user-facing launchers now open it. Registration is complete on all four surfaces, and the idle-reset block is safe on the first present because `window_start` gates the subtraction.

shibco:
"Should" fix --_--

ClickSentinel:
> "Should" fix --_--

gonna create an issue for the attitude

shibco:
did fix
what's the issue

ClickSentinel:
Reviewed at `ac088f5`, covering both commits since the last posted pass at `0f57214`. Method: full container build and audit, patch-check, the counter logic re-modelled across all three revisions, and the launcher blocks run as shell replicas with a fake Live process in the process table. No Live session — the shell changes are settled by execution and the counter changes by the model that produced the last pass's finding.

**Everything from the last pass is fixed, and the two that were measurable were re-measured.**

The idle-reset finding: same model, one class per revision, same scenario table. An oscillating fault in 8 s bursts separated by 6 s pauses was silent for the whole 80 s / 4801-of-4801-frame run under `0f57214`; under `ac088f5` it warns at 19.0 s. The persistent shape still fires at 2.0 s and the healthy 30% resize-burst session is still silent. The evidence-line finding: the 55%-drag case went from `[consec 0 pair 0]` to `[consec 9 pair 1]`, so the zeroed report is gone. The tester kit has `WINEDEBUG` in its `unset` list again, and `max9` and `ableton-wine-portal` now default to `-all,+winediag`.

The `journalctl` point is answered properly. Replicating the block and `exec`ing into `wine wineboot -u` on a throwaway prefix, Wine's own stderr — not just the script's `echo`s — crosses the exec and lands in the file, with the terminal copy identical. The `.als`, `.auz` and `ableton://` entries all do `Exec=…/ableton-live`, so the `pgrep` guard's premise is right, and with a matching process in the table the re-launch appends instead of truncating.

**Should fix**

- **A refused launch truncates the log it was meant to preserve.** The truncation sits above every early-exit path in both launchers. In `bin/ableton-live-beta` the block is lines 12-18 and the "Suite is running" refusal is lines 33-36, so a tester who has stable Live open loses the previous beta session's `live-beta.log` and is left with only the refusal message — measured with a replica and a fake Suite process. `scripts/ableton-live` has the same shape against its single-instance guard: with Live not yet in the process table, a second launcher truncates the first one's log and is then refused at line 153. Fix: rotate instead of truncating (`mv live.log live.log.1` beside the existing append), or move the truncation below the guards that can refuse.

**Comment only**

- **The disclosed false-positive window is 10 s narrower than the code's.** The notes give the edge as "a 20-second gap" and contrast 20 s against 60 s. Sweeping the gap on two 6 s bursts, the warning fires at every gap through 29 s and goes silent from 30 s — the edge is `PRESENT_MISMATCH_STRIKE_RESET_MS` itself, which is also the more defensible way to state it. In the same paragraph, "5.5 seconds and longer warned after 16.5 to 19 seconds" holds for bursts of 5.5 s to 8 s; a 10 s burst warns at 21 s. Fix: "any gap shorter than 30 seconds", and drop the upper bound on the burst range.

- **TROUBLESHOOTING names one log path and the beta writes another.** The section tells every reader to grep `~/.log/ableton-wine/live.log`, but `bin/ableton-live-beta` tees to `live-beta.log`. Beta testers are the population most likely to be asked for this line. Fix: one clause naming the beta path.

- **`SERIES_GAPS[0057]` is stale — pre-existing on main, not this PR.** `0057-wined3d-add-Intel-graphics-devices-from-Ice-Lake-to-.patch` is in the series and fingerprinted at `build-audit.sh:136`, while line 74 still lists 0057 as reserved. The numbering loop only consults the array for numbers that are missing, so it passes silently — the same hazard this PR's own 0070 entry will become. Worth sweeping the array while 0070 is being added.

**Held**

Registration is complete on all four surfaces: the `SERIES.sha256` checksum matches the rewritten patch, `BASE.txt` describes 0071 and the 0070 reservation with the count right at 65, `build-audit.sh` fingerprints it, and the notes carry the companion doc. patch-check is clean and the series applies to `base-5c23dd1c`; the three "no blank line" formatting notes are pre-existing on 0008, 0009 and 0054. The build audit passed 94 checks with 0071 `PASS sha✓ stamp✓` and the string present in the shipped `wined3d.dll`, and the merge against the declared base `main` is clean.

The two must-fixes from the first pass still hold at this revision. The destroy summary is gated on `warned`, and the summary text is `Swapchain %p present-size mismatch summary:`, which the TROUBLESHOOTING grep cannot match — so check 3 can now pass as written. The ratio block's brace structure matches what the model assumes: `window_strikes` is zeroed in the `else`, not unconditionally, so a strike genuinely survives to the second window. `ableton-wine-portal` is not a Live launch path — only `bin/set-file-portal-policy` calls it — so it needs no tee.

The "Status on 2026-08-08" paragraph is the right call: those figures are model output, and saying so is better than letting them read as session data.

PR #144, opened 2026-08-05
Bring the installer improvements to Live 11 install

shibco:
Adds additional work to the Live 11 installer to remove freezes with the USB driver installer, etc

### DIFF
--- a/TROUBLESHOOTING.md
+++ b/TROUBLESHOOTING.md
@@ -11,8 +11,8 @@ window. The installer then continues by itself. An update that stops at
 
 Ableton's own installer adds a small Windows helper program that Live does
 not need on Linux. The helper stays open, and setup waits for it. Releases
-newer than 2026.08.04.1 skip the helper and close leftover copies, so the
-stop does not come back.
+newer than 2026.08.04.1 close leftover copies of the helper, so the stop
+does not come back.
 
 To update, download
 [the latest installer](https://github.com/shibco/ableton-linux/releases/latest/download/install-ableton-latest.run)
--- a/TROUBLESHOOTING.md
+++ b/TROUBLESHOOTING.md
@@ -3,6 +3,27 @@
 Here are some common ailments we've seen, and how to fix them.
 
 
+## The installer does not finish after Live installs
+
+Look for an **Ableton USB Driver** window in your taskbar. Close that
+window. The installer then continues by itself. An update that stops at
+`== [1/5] initialise prefix ==` has the same fix.
+
+Ableton's own installer adds a small Windows helper program that Live does
+not need on Linux. The helper stays open, and setup waits for it. Releases
+newer than 2026.08.04.1 skip the helper and close leftover copies, so the
+stop does not come back.
+
+To update, download
+[the latest installer](https://github.com/shibco/ableton-linux/releases/latest/download/install-ableton-latest.run)
+and run:
+
+```bash
+sh ~/Downloads/install-ableton-latest.run --update
+```
+
+The update keeps your Live installation, your license, and your projects.
+
 ## Live has no sound
 
 Open **Settings > Audio** and select:
--- a/TROUBLESHOOTING.md
+++ b/TROUBLESHOOTING.md
@@ -5,16 +5,9 @@ Here are some common ailments we've seen, and how to fix them.
 
 ## The installer does not finish after Live installs
 
-Look for an **Ableton USB Driver** window in your taskbar. Close that
-window. The installer then continues by itself. An update that stops at
-`== [1/5] initialise prefix ==` has the same fix.
-
-Ableton's own installer adds a small Windows helper program that Live does
-not need on Linux. The helper stays open, and setup waits for it. Releases
-newer than 2026.08.04.1 close leftover copies of the helper, so the stop
-does not come back.
-
-To update, download
+If an **Ableton USB Driver** window is in your taskbar, close it. The
+installer then continues by itself. If there is no such window, press
+Ctrl-C, then download
 [the latest installer](https://github.com/shibco/ableton-linux/releases/latest/download/install-ableton-latest.run)
 and run:
 
@@ -22,7 +15,14 @@ and run:
 sh ~/Downloads/install-ableton-latest.run --update
 ```
 
+An update that stops at `== [1/5] initialise prefix ==` has the same fix.
 The update keeps your Live installation, your license, and your projects.
+
+Ableton's own installer adds a small Windows helper program that Live does
+not need on Linux. The helper stays open, often without any window, and
+setup used to wait for it. Releases newer than 2026.08.04.1 stop the
+helper themselves and remove its autostart entry, so the wait clears after
+about half a minute and the stop does not come back.
 
 ## Live has no sound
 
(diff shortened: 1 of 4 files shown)


ClickSentinel (changes requested):
I ran this against a real Live 11 Suite 11.3.25 install with the full payload, on the current release runtime (`wine-d2d1-nspa-11.13`, dist 2026.08.04.1, 65 patches). I also pulled the bundle's Burn manifest out of the installer and read it directly, which is where most of the below comes from — those bits don't depend on the Wine build at all. I ran the whole thing twice on two different runtimes and got identical results.

Short version: the Suite bundle isn't put together the way the notes describe, so on Suite the seed doesn't do anything.

**There's no guard to satisfy.** The fix hangs on the bundle checking `InstalledPush3AudioDriverVersion <= v5.68.0`. That check isn't in this bundle. There's exactly one `MsiProductSearch` in the whole manifest and it's for the VC redist. There's no `InstallAudioDriver` variable either. The driver package's only condition is `InstallCondition="VersionNT64"` — on 64-bit, it installs. Burn's log agrees: it planned the package in, ran it, and executed `tlsetupfx.exe`.

| | 11.3.35 trial (notes) | 11.3.25 Suite (measured) |
| --- | --- | --- |
| Driver package | `Push3AudioDriver` | `Ableton_Push_Audio_Driver.msi` |
| Version | 5.68.0 | 5.57.0 |
| Install condition | `… InstallAudioDriver > 0 And InstalledPush3AudioDriverVersion <= v5.68.0` | `VersionNT64` |
| `MsiProductSearch` on driver UpgradeCode | present | absent |

**The end result on Suite is just the old behaviour.** The driver package runs and fails the same way it always did:

`w350: Applied non-vital package: Ableton_Push_Audio_Driver.msi, encountered error: 0x80070643. Continuing...`

That's the same `0x80070643` your comment at `setup-prefix.sh:634` already describes. It's marked `Vital="no"`, so Burn shrugs and carries on, and Live installs fine. Same as before this PR.

**The placeholder gets removed.** The UpgradeCode you seed is in the manifest, but as something Burn cleans up: `<RelatedPackage Id="{AEFC5C68-0264-4E30-9685-28712A91CF4E}" MinVersion="0" MinInclusive="yes" OnlyDetect="no" />`. With `OnlyDetect="no"`, Burn treats anything registered under that code as an old product to uninstall. That's what happens — I confirmed the seed reads back as `99.0.0` right up until the install, and afterwards the `UpgradeCodes` entry is gone, leaving only the orphaned `InstallProperties` key. Both runs, both runtimes.

**`/passive` still needs clicks, and that's the bundle's call rather than Wine's.** `Setup.msi` is marked `DisplayInternalUI="yes"`, so its wizard opens even in passive mode — it would do the same on Windows. I had to click through several dialogs both times. That makes the "a progress window opens, no clicks needed" message wrong for this bundle.

**Plenty here does work.** Detection is spot on — `.wixburn` sits at byte 632, well inside the 4k window, exactly as your comment says. The seed mechanism itself is sound: I wrote a small probe calling the same pair of functions `MsiProductSearch` uses, and in a seeded prefix it returns `rc=0 product={B0B57A61-…}` with `VersionString 99.0.0`, from both 32- and 64-bit callers, so a 32-bit Burn stub would see it fine. The `taskkill` does its job too — a parked process blocks `wineserver -w` (rc=124) and `taskkill /f /im` clears it.

**One small thing.** The scrub misses `HKCU\Run` and the Startup folder, which `wineboot -u` also reads (`wineboot.c:1931`). I checked: the `HKCU\Run` value survives and the tray app comes back every boot. No hang results, though — pulling the two halves apart, `taskkill` alone is clean for both hives while `reg delete` alone hangs for both, so the taskkill is doing all the real work. More a tidiness thing, but "wherever the driver registered it" claims a bit more than the loop covers.

**One question for you.** Does the 11.3.35 trial bundle definitely have `InstallAudioDriver` and the version guard? If so, trial and suite genuinely differ and this is a trial-only fix. And if the trial carries that same `RelatedPackage` removal entry, it's worth double-checking what actually suppressed the driver in your testing, since the seed would be getting swept there too.

ClickSentinel:
Second pass at `e25baf29`, driver and installer level: ran the new blocks directly against a scratch prefix and the detection gate against real Live 11 bundles of both generations. No Wine build, no Live install.

Everything from the last round is addressed, and I re-ran the harnesses rather than reading the diff. Gate correct on four stubs (suite 11.3.25 no, trial 11.3.25 no, trial 11.3.35 yes, trial 11.3.42 yes). Seed survived the move — probe reads back `rc=0 {B0B57A61-…} VersionString 99.0.0`. Cleanup clears all four registrations. Run scrub takes all three hives with a spaced value name (1,1,1 → 0,0,0). The `find` guard lands: old form exited 1 and never reached `wineserver -w`, new form reaches it.

**Should fix**

- **The wait #111 reported is still unbounded.** #111 hung at `[1/5] initialise prefix` — `setup-prefix.sh`'s `"$WINESERVER" -w`. The 30 s bound went on the other path, in `setup-run-header.sh`, so the reported symptom still rests entirely on the taskkill above it landing. Fix: bound that one too.

- **The taskkill's image name has never been verified — my correction as much as anyone's.** The harness I used last round created a stand-in *named* `tusbaudiocplapp.exe`, so it only proved the scrub kills `tusbaudiocplapp.exe`. The driver MSI log from that same session shows the Live 11 package ("Ableton Push Driver v5.57.0") installing `AbletonPushCpl.exe` into `C:\Program Files\Ableton\Push Driver\x64\` with a `StartUp\` shortcut, and no file named `tusbaudio*` — the only such string is the log filename. Different package from Live 12's "Ableton USB Driver" (`audiodriver_x64.msi`), which is what the reporter had, so the name may be right there; it is not right for the Live 11 driver. Fix: cover both names, or match Run and Startup entries by their target path under `Ableton\`, and confirm the Live 12 name from `audiodriver_x64.msi`'s File table.

- **The 30 s fallback re-enters an unbounded wait.** After the bound it calls `wineserver -w` again with no bound, so the Live 12 hang is unchanged — that one parks on `MicrosoftEdgeUpdate.exe`, which has no window, and the new message asks the user to close one. Still better than silence, but it reads as handled. Fix: after the bound, kill residents matching `WINEPREFIX` in `/proc/<pid>/environ` rather than by name, or carry on without waiting; drop the claim that a window exists.

- **The case-sensitivity comment names the wrong invariant.** `-i` changes nothing on any of the four stubs — WiX 3 carries `WiX Toolset Bootstrapper`, spaced, so a case-insensitive `wixtoolset` still misses it. The missing space protects the gate, not the case, and the real fragility is someone allowing a separator. Fix: reword to name the space.

**Comment only**

- `setup-prefix.sh`'s "installed Live before 2026.08.04.1" is off by one: that release is current and lacks this fix, so its prefixes have the tray app too. TROUBLESHOOTING's "newer than" is correct. Fix: name the release carrying the fix.

- The notes' open item is the right one to keep visible — seed verified on trial 11.3.35 only, suite 11.3.42 untested.

**Held**

Both packed GUIDs are correct MSI-packed forms. Ordering right: `setup-prefix.sh` at run-header line 319, seed at 372, so cleanup cannot wipe the seed it is about to write; `--refresh` does not short-circuit before the cleanup, so `--update` clears leftovers. Dropping `2>/dev/null` from `head -c 4M` costs nothing — a SIGPIPE'd `head` prints nothing — and surfaces a read error that was silent. Merges clean into main at `d8d293c`. `shellcheck -S warning` adds nothing new; the SC30xx noise in `setup-run-header.sh` is pre-existing.

shibco:
feedback addressed

PR #145, opened 2026-08-05
Performance Moonshot: broader CPU fixes  

shibco:
This PR is part of the Performance Moonshot series.

Live's audio engine and its interface exchange many small messages all the time, also when nothing plays. Wine processed each message through the wineserver and when this gets out of control it causes unpredictable CPU usage with negative effects on performance. Previously we were relying only on `ntsync` for this, and that's fine, but there are some instances where it's not fine. This patch will remove most of churn.

This wine patch is on by default and can be disabled via

```
WINE_MSG_FASTPATH=off ableton-live. 
```

One delivery-order rule for thread callbacks is relaxed on purpose; `tools/apcprobe.c` documents and checks it. 

Technical detail is in the patch messages under `patches/performance/`.


ClickSentinel (changes requested):
Reviewed at 1b92d9d. Method: read the five patches against the 11.13 base sources, ran patch-check, built the runtime, ran build-audit, and exercised 0002/0005 with headless PE probes against the build in an isolated prefix. Live was not launched.

patch-check is clean (67 apply, 72 manifest entries) and the build audit reproduces the findings doc exactly — 95 checks, every patch verified, all five fingerprints present. Three blocking findings below; a handful of should-fix and doc items to follow separately.

### 1. `performance/0002` — APC handle cache has no invalidation on close, so APCs are delivered to the wrong thread

`resolve_client_apc_tid` keys `{HANDLE -> tid, generation}` on the handle value alone. The generation guards *tid* recycling; nothing guards *handle* recycling. `alloc_entry` (`server/handle.c`) hands a freed slot straight back, so a closed thread handle whose value is reused for a different live thread still validates, and the APC goes to the original thread.

Reproduced — the handle came back on the **first** `CreateThread` after `CloseHandle`:

```
step1  queued via hA=0x38 -> ran on tid 288 (A=288)  ok
step2  handle 0x38 reused for a NEW thread, tid 292
step3  queued via reused handle -> ran on tid 288
RESULT *** MISDELIVERED to the old thread ***
```

Same probe with `WINE_APC_FASTPATH=off` delivers to tid 308, correctly. `CloseHandle(CreateThread(...))` fire-and-forget makes this ordinary rather than exotic.

**Fix:** invalidate matching cache entries from `NtClose`, or drop the handle cache and resolve per call.

### 2. `performance/0002` — `apc_mutex` is taken with signals unblocked, and the SIGUSR1 handler re-enters it

`usr1_handler` -> `wait_suspend` (`thread.c:1558`) -> `server_select` -> `apc_fastpath_enter_unarmed_wait()` -> `mutex_lock(&apc_mutex)`, unconditionally — while `queue_client_apc` and `apc_fastpath_fetch_client_apc` hold that same non-recursive mutex across an ntsync ioctl. A SIGUSR1 in that window self-deadlocks the thread and leaves `apc_mutex` held for the life of the process.

Reproduced, three arms of one probe (queuer thread, plus a `SuspendThread`/`ResumeThread` loop):

| arm | round 1 | round 20 |
|---|---|---|
| fastpath on, with suspend | queued 14656 | queued **14656** — wedged |
| fastpath off, with suspend | queued 15161 | queued 19772 |
| fastpath on, no suspend | queued 100476 | queued 1755629 |

Both the queuer and the APC target stop permanently; only the thread that never touches the APC path keeps running. Exposure is wider than `SuspendThread` — `thread_queue_apc` (`server/thread.c:1475`) also signals SIGUSR1 for a system APC queued to a thread not already in an APC wait, i.e. ordinary async I/O completion.

**Fix:** block `server_block_set` around every `apc_mutex` section, the way `virtual.c` uses `server_enter_uninterrupted_section` for `virtual_mutex`.

### 3. `performance/0005` — the known-clean `GetUpdateRect` stops writing `*rect`

The patch message says "today the rectangle is left untouched as well". It is not: `DECL_HANDLER(get_update_region)` succeeds with zero rects, `NtGdiExtCreateRegion` returns a *valid* empty region (`alloc_region` clamps to `RGN_DEFAULT_RECTS`), `send_ncpaint` returns it non-NULL, and `NtGdiGetRgnBox` writes `{0,0,0,0}` into `rect` before returning NULLREGION. That is also the documented Windows behaviour.

Reproduced with a poisoned RECT:

```
fastpath on   clean ret=0 rect={1515870810,...}   <-- untouched
fastpath off  clean ret=0 rect={0,0,0,0}
```

**Fix:** `if (rect) SetRectEmpty( rect );` before the early `return FALSE`.

ClickSentinel (approved):
Runtime CPU at `9504981`, since the thread has driver-level probes but no end-to-end numbers. PR build against a cloned licensed prefix, Live 12.4.3 idle, PipeASIO; 3×60s per arm in one launch, `utime+stime` from `/proc`, paired by sample index (both arms drift down 1→3 as Live settles).

| component | arm | s1 | s2 | s3 | mean | change |
|---|---|---|---|---|---|---|
| **wineserver** | on | 1.75 | 1.75 | 1.77 | **1.76** | **−16.7%** |
| | off | 2.40 | 2.02 | 1.91 | **2.11** | |
| **Live main** | on | 12.13 | 12.00 | 11.68 | **11.94** | none |
| | off | 11.94 | 12.15 | 12.00 | **12.03** | |
| **total** | on | 14.60 | 14.19 | 13.78 | **14.19** | **−3.0%** |
| | off | 15.01 | 14.61 | 14.29 | **14.64** | |

cpu-seconds per 60s window. wineserver arms do not overlap (on max 1.77 < off min 1.91, p≈0.05 by permutation); Live's own process shows no effect and inconsistent paired signs. Mechanism confirmed, landing where it should.

Two doc reconciliations, no code defect:

- P5's outcome column reads "30-40% idle core to under 5%, server churn cut". Churn is cut; idle total measured 23% of a core, 20 points of it inside Live's own process, so "under 5%" is not reachable from the server side. P13 already records "main GUI thread is the hottest thread". *Fix: scope P5's outcome to server churn, let P13 carry the idle-core figure.*
- CHANGELOG pairs "Live uses less CPU when nothing plays" with ">99%" a few lines below — that 99% is request counts, the felt CPU is ~3%. *Fix: bind the 99% to request counts explicitly.*

Well done.

shibco:
Hell yes. I'm merging and will complete an inline documentation pass after we are done with this round of PRs

PR #148, opened 2026-08-05
Better Desktop Launchers

shibco:
Outside of @yioannides's amazing iconography work, I really fucked up the desktop launcher script. I've cleaned this up, and they should work better than they did.

Thanks @jackson-57 for the sharp eye in spotting this.


ClickSentinel:
My AI overview said there is a dead icon path in the beta template. points `Icon=` at `~/.local/share/icons/hicolor/256x256/apps/ableton-live-beta.png`. Nothing in the repo ships that file and no script installs it, can you confirm?

PR #155, opened 2026-08-07
𝓑𝓮𝓪𝓾𝓽𝓲𝓯𝓾𝓵 𝓯𝓸𝓷𝓽𝓼 𝓲𝓷 𝓦𝓲𝓷𝓮

shibco:
Lol, a bit psyched about this. This branch addresses something that had driven me *absolutely crazy* for 10 years, and especially since switching to linux full time in 2020. Adds patches 0081 through 0089 and three probes to bring ClearType-style font rendering to Wine. 

The whole thing is implemented via DirectWrite rasterising ClearType with three samples per pixel, Direct2D handling each channel, and forcing GDI to set antialiasing from the prefix instead of the host.

Experimental! Probably OK to put in a release! BUT EXPERIMENTAL! Hello Wine devs, I want to upstream this to you, please let me!


shibco:
whoops lol try again @ClickSentinel

ClickSentinel:
Reviewed at `ea4db4c0` by reading the nine patches against `base-5c23dd1c`, then building and measuring three runtimes side by side: `A` = upstream main + PR 157 (none of 0081–0089), `B` = `615a4c8`, `C` = `ea4db4c0`, each identified by the `0087` line of its `ABLETON-WINE-PATCH-STACK.txt`. DirectWrite and Direct2D driven headless with the PR's own probes plus three variants. Live was never launched, so nothing below speaks to Live's own render target.

`patch-check.sh` clean, `build-audit.sh` 112 checks, merge with `upstream/main` clean.

## Blocking

None.

## Should fix

- **`tools/cleartype-probe.c` reports `GREYSCALE` on a working build.** Its verdict is `ink[0] == ink[1] && ink[1] == ink[2]` — channel *sums*, equal for any symmetric filter, and 0081's `{0x08,0x4d,0x56,0x4d,0x08}` is symmetric. Its default em size of 16 also sits inside the embedded-bitmap range of Wine's bundled `tahoma.ttf` (`EBDT`/`EBLC`), where coverage is a flat 255 and `freetype_get_lcd_glyph_bitmap` never runs. Measured per sample at em 24: baseline 0 differing triples of 2131, `ea4db4c0` 2573 of 2676, texture two pixels wider — 0085's filter-tail padding. Fix: compare per triple, and default outside the strike range.
- **The launcher writes over an explicit `FontSmoothing=0`.** The head commit made smoothing-off a deliberate choice that outranks the host; `sync_text_smoothing` rewrites it to `2` on the next launch, because the guard only returns early when the prefix already holds the *wanted* pair. `ABLETON_TEXT_SMOOTHING=preserve` is the only escape. Fix: leave a prefix that says `0` alone.
- **`ABLETON_TEXT_SMOOTHING=grayscale` governs Direct2D only.** It writes `FontSmoothingType=1`, which the head commit defines as recording nothing the user chose, so GDI falls through to the host. Measured on `ea4db4c0`: Direct2D 0.000 spread at em 32, GDI identical to `subpixel` at every size 9–32. Fix: give greyscale a state 0087 honours, or scope the comment to Direct2D.
- **`dist/BUILD-INFO.txt` and `dist/BUILD-INFO-2026.08.04.1.txt` understate the series by nine patches.** Committed 65 / 63 with `patch-stack: 9e48edd6…`; building this tree rewrites both to 75 / 73 with `patch-stack: 7d008967…` and `dist-version` unchanged, so one version string names two stacks. Inherited drift (main is 64 against the same 63) taken from one to ten. Fix: regenerate alongside the version bump.
- **`BASE.txt` says numbers 0066 through 0080 are reserved**, but 0069 is in the series. `SERIES_GAPS` correctly omits it, so only the prose is wrong. Fix: "0066 through 0068 and 0070 through 0080".
- **0088 is unreachable and declines an available fingerprint.** Nothing in the repo writes `HKCU\Software\Wine\Fonts\DesktopUIFont`, and its audit entry claims "adds no string literal" while `strings win32u.so` finds it once here and zero times on the baseline. Fix: a `FINGERPRINTS` line, and a sentence in `BASE.txt` saying it is for downstream use.

## Comment only

- **ClearType is unreachable at Win32 UI sizes with Wine's own Tahoma** — flat 255 coverage at em 12 and 16, outline path live from 24. Live's Ableton Sans faces are outline-only, but Wine-drawn dialogs at 11–16 px get nothing.
- **The three-pass blend costs about double, not triple.** 2000 `DrawText` calls, 24-glyph run at em 32, median of three: 1028 ms against 502, 515 and 531 ms on three greyscale controls.
- **0086's opaque-target restriction verified by execution** — the same probe with `D2D1_ALPHA_MODE_PREMULTIPLIED` returns 0.000 where `IGNORE` returns 53.991. `tools/d2d-text-probe.c` builds its target as `IGNORE`, so it passes the gate by construction and cannot show a declined target.
- **0084's premise does not hold on a fontconfig-subpixel host.** win32u already resolves subpixel per font — `FC_RGBA_RGB → WINE_GGO_HRGB_BITMAP` at `dlls/win32u/freetype.c:1162`, read back as `font->aa_flags` at `font.c:2471` — so the baseline arm renders GDI text subpixel in every prefix state. This box also has no `Xft.*` resources, so `WINE_GGO_AA_FROM_HOST` is never applied and gate-on/gate-off with `WINE_DISABLE_PREFIX_FONT_SMOOTHING=1` were indistinguishable for that reason, not because the switch is inert.
- **0086 fetches `gamma` and `contrast` and uses neither** outside the TRACE, so the blend is not gamma-corrected. Consistent with the greyscale path; worth a line in `BASE.txt` given the upstreaming intent.
- **No `notes/` companion** for a nine-patch, 1900-line subsystem. `BASE.txt`'s per-patch entries are unusually good, but 0031 and 0032 set the precedent.

## Held

0089 is correct — `unsigned int gray` made `pixel[sample] - gray` wrap and saturate a channel to 255. 0085's `has_contours` is the right predicate: `freetype_glyph_has_contours` is `format == FT_GLYPH_FORMAT_OUTLINE && outline.n_contours` (`dlls/dwrite/freetype.c:111`), the same condition `freetype_get_lcd_glyph_bitmap` branches on. `WINE_GGO_AA_FROM_HOST` cannot escape to `GetGlyphOutline`: every branch under `if (font)` assigns `*aa_flags`, and a NULL font returns 0. `dibdrv_SelectFont`'s `<= 8bpp → GGO_BITMAP` constraint survives 0087's rewrite through `if (incoming && !host_default)`. 0083 needs no Makefile change — dwrite already imports `user32`.

giang17:
Hi Cade,

we picked up the nine ClearType patches from this PR into `giang17/wine`
(`d2d1-dcomp-11.0`) with your authorship intact, and they have been running here
on Fender Studio Pro 8, FL Studio and Ableton Live 12 for a day. Thank you — the
difference is visible, and our maintainer (a graphics designer by trade) describes
the result as reading "like print". Below is what came back out of testing them,
in case any of it is useful for the WineHQ submission.

Everything here is measured on `5c23dd1c` plus your series, X11, NVIDIA, 92 dpi,
`hintslight`, `XftSubPixel=rgb`.

---

## 1. A gap in 0081: embedded bitmap strikes defeat it for the default UI font

This is the one we think matters most for upstreaming.

A ClearType texture needs three different coverage samples per pixel, and only an
outline can supply them. Where a face carries an embedded bitmap strike, FreeType
returns the strike, and `glyphrunanalysis_render()` expands its single sample into
three identical subpixels — the exact loss 0081 sets out to remove, reintroduced
one layer lower.

It is not a corner case: Wine's bundled `tahoma.ttf` carries strikes for 8 to 16
ppem, and `MS Shell Dlg` resolves to Tahoma. So the default interface font lost its
subpixel resolution across the whole size range interface text actually uses. The
strike being 1bpp, those glyphs were not even greyscale antialiased.

Fix: ask for the outline when the caller wants a ClearType texture. A face with no
outline for the glyph keeps its strike, so the worst case is the result you would
have had anyway; the greyscale and aliased paths are untouched. The bounding box
goes through the same helper and carries the same flag, so box and rasteriser
cannot disagree about which form was used — which in turn makes the form part of
the glyph cache key, alongside the `lcd` flag you already added there.

Measured with a D2D test program on an opaque target, counting pixels whose
channels differ by more than 15 of 255:

| | before | after |
|---|---|---|
| MS Shell Dlg 11px | 0.00% | **11.50%** |
| BarlowSemiCondensed 10–13px | 0.00% | **8.86%** |
| GRAYSCALE and ALIASED control lines | 0.00% | 0.00% |
| 16px outline fonts | unchanged | unchanged |

Fender Studio Pro 8 renders bit for bit identically, its faces having had outlines
all along.

The change is here, if you want to look at or cherry-pick it:
**[giang17/wine@`21a3723`](https://github.com/giang17/wine/commit/21a37233cbc2ac3aaa8ffc495f2e5bdccfb81650)**
— five files in `dlls/dwrite`, 43 insertions.

It is written against our tree (`d2d1-dcomp-11.0`), so it will likely need a rebase
onto yours. Happy to do that and open it as a PR in your series format — just say
which number and base you want it on, since I would rather not guess at your
numbering and `SERIES.sha256` handling.

---

## 2. The ~2x cost figure: both measurements are right

@ClickSentinel measured the three-pass blend at roughly double. We measured 1.04x.
Both are correct, and the difference is the harness — worth knowing before the
number ends up in a review thread as an objection.

| | extra cost per `DrawText` | ratio | baseline per call |
|---|---|---|---|
| @ClickSentinel | 256 µs | 1.99x | 258 µs |
| ours (`ID2D1HwndRenderTarget`) | 173–196 µs | 1.04x | 3474 µs |

The **absolute** cost of the extra passes agrees within the spread — our own runs
reach 253 µs. What differs is the baseline: their harness spends 258 µs per call,
ours 3474 µs, a factor of 13. The same added work lands as +99% there and +4% here.

So the portable statement is the absolute one: **each ClearType glyph run costs
roughly 0.2 ms more per `DrawText` than the greyscale path**, and how much that
matters depends entirely on what the application does per call besides drawing
text. For an application with a real window render target — which is what plugins
have — it is a few percent.

Method: 2000 `DrawText` in one `BeginDraw`/`EndDraw`, best of 5 after warmup,
three independent program runs, real GPU, `WINEDEBUG=-all`. Fixed-text axis is the
trustworthy one (three runs within 12%); the varying-text axis scatters and should
not be quoted alone.

---

## 3. A measurement pitfall that cost us half a day

Flagging this because it is easy to hit and it silently produces the wrong
conclusion. When counting colour fringing by channel spread, ImageMagick's `txt:`
output here is **8-bit**, not 16-bit. A snippet that divides by 257 scales every
real difference to under 1 and reports "0.00% fringing" for a perfectly good
ClearType build.

We concluded from that the patches only produced "finer greyscale" and wrote it
into our documentation, before re-measuring showed 29% of body-text pixels in
Studio Pro carrying a fringe. Reading the maximum from the `txt:` header line
instead of assuming a quantum depth fixes it, and always measuring a GRAYSCALE or
ALIASED control line — which must stay at exactly 0.00% — catches it immediately.

Related, and consistent with @ClickSentinel's finding about `cleartype-probe.c`:
any metric that collapses the three channels (greyscale conversion, channel sums)
is blind to precisely the thing being measured.

---

## 4. Where 0084/0087 land on a fontconfig-subpixel host

@ClickSentinel's prediction held exactly. On this machine the GDI half changes
**0 of 4000 pixels** in the Win32 menu bar, because win32u already resolves
subpixel per font (`FC_RGBA_RGB → WINE_GGO_HRGB_BITMAP`) and the host reports
`rgba: 1`. That is the default state on KDE and GNOME, so the prefix-precedence
work has no effect for a large share of users — worth knowing when defending those
two patches upstream, since the reviewer may well be on such a host.

---

## 5. Two things we added on top, in case they are interesting

Neither is a defect in your series; both are opt-in and default to today's
behaviour bit-identically.

- **Enhanced contrast is fetched and dropped.** `GetAlphaBlendParams()` reports
  gamma, contrast and level; only the level is used. We apply the contrast as
  well (raising partial coverage, endpoints fixed, monotonic) behind
  `HKCU\Software\Wine\Direct2D` → `text_enhanced_contrast`
  ([`8a3a716`](https://github.com/giang17/wine/commit/8a3a71682dfc0b705cdeff8e5c2b736dfa17d985)).
  Wine's dwrite hardcodes
  the reported value to 0, so honouring it changes nothing until the key is set.
  On a dark interface at 0.70 our maintainer reports noticeably better legibility —
  light text on dark reads thinner than it is (irradiation), and dense body text
  suffers most. Audio software being what it is, that is the common case.
- **Strike vs outline is selected on the wrong axis.** Windows decides by
  *rendering mode* — GDI modes take the strike, NATURAL modes the outline — while
  we (and, after 0081, you) decide by whether ClearType was asked for. Consequence:
  greyscale text still takes the strike, and since Tahoma has no strike for *every*
  size, the character flips between neighbours — 12, 15 and 16px come out as hard
  1bpp while 14px and 17px+ are antialiased. Both applications we instrumented
  request `NATURAL` exclusively (Studio Pro: 1625 of 1625 glyph-run analyses), so
  the strike is being taken in exactly the modes where Windows would not take it.
  Ours is behind `HKCU\Software\Wine\DirectWrite` → `outline_in_natural_modes`
  ([`2a00f9b`](https://github.com/giang17/wine/commit/2a00f9b7a51a7779db85e5b7fe24b633cdfa73fe)).

---

Also: `gamma` and `contrast` are fetched in `d2d1` and only `cleartype_level` is
used, so the blend is not gamma-corrected. We looked at whether to apply a curve to
the coverage and decided against it — correct handling needs the blend to happen in
linear space (sRGB targets or an explicit linearisation in the shader), and a fixed
curve improves one polarity at the expense of the other. Filed on our side as a
feature rather than a defect, since nobody has complained about how it looks.

Thanks again for the series, and to @ClickSentinel for the review — the
`cleartype-probe.c` verdict and the 0084 premise were both called correctly, and
both matched what we saw here.

PR #156, opened 2026-08-07
Slayed SIX Win32 application compatibility faults

shibco:
Adds patches 0075 through 0080: three missing exports that aborted the calling process, the upstream fix for the 11.13 window sizing regression, Motif hints for disabled windows, and a dxgi surface search that ran on every ReleaseDC.

This is not directly related to Ableton Live, and I am ABSOLUTELY ENJOYING MYSELF merging this particular commit, knowing what it means for the future of this project  :shaking_face:


ClickSentinel (approved):
Read all six patches against `base-5c23dd1c`, ran `patch-check.sh`, built the branch and read the audit, merged against `main`, then probed the built runtime for each patch with the installed runtime (main @ 0065) as the control arm. Live was not in the loop.

Nothing blocking. Build clean, audit 103/103, series applies. All six fixes verified working: 0078 sizes an 800×600 request correctly at 120 DPI where the control gives 512×384; 0077 moves the Motif functions 0x26 → 0x3e on a `WS_DISABLED` top-level, matching your trace; 0075/0080 no longer abort with `EXCEPTION_WINE_STUB`; 0076 matches an independent SHA-256 on three names; 0079 drops 40 blits per 20 frames and, worth noting, also stops an unrelated same-sized popup being overwritten — that's a correctness fix, not only cost.

pls fix:
- `0080` FIXMEs on every call to `ProcessPointerFramesInteractionContext`, the one function here that runs per frame — measured 8 calls, 8 lines under Wine's defaults. The launcher's `WINEDEBUG=-all` shields shipped sessions, direct runs get a line per frame. Fix: `if (!once++)` on that one only; suggestions below carry the hunk-header and `SERIES.sha256` companions.

minor:

- `0080` adds five exports without declarations in `include/interactioncontext.h`, while 0076 adds its own to `include/userenv.h`. Consistency only — Wine's warning set here has no `-Wmissing-declarations`.
- `build-audit.sh:74` `SERIES_GAPS[0057]` is stale; 0057 is in the series and fingerprinted on line 141. Pre-existing on main — the numbering loop only reads that array for genuinely missing numbers, so 103/103 PASS says nothing about it.
- `STAMP_ONLY` entries for 0077/0078/0079 sit before `0002`; the rest is numeric.

ClickSentinel:
Second pass at `2165d9b`, covering both commits since the posted pass at `d4ef246`. Method: full container build and audit, patch-check, declared-base merge, and the ninput probe re-run across both revisions. Patch and runtime level — no Live session.

**Every posted finding is fixed, and the measurable one was re-measured.**

The once-only guard, measured rather than read. `probe_fixme_rate.c` calls two exports four times each in one process; the sibling export is the control, since 0080 gives `AddPointerInteractionContext` a per-call FIXME and `ProcessPointerFrames` a once-only one — both arms, same binary, same debug channels. `d4ef246`: 4 and 4. `2165d9b`: **1** and 4.

The five `interactioncontext.h` declarations match the implementations and the spec arities; `ninput/tests/ninput.c` has no conflicting declaration and the header has no consumer outside ninput. `SERIES_GAPS[0057]` is gone and the gap list is now exactly the ten missing numbers. `STAMP_ONLY` is numerically ordered.

**Should fix**

- **BASE.txt says 0078 does not carry the protocol bump; it does.** `patches/BASE.txt:273-275` says the `SERVER_PROTOCOL_VERSION` bump "rides the performance series patch that follows in this stack". 0078 raises it itself, 956 → 957, and is the only patch in the series touching `server_protocol.h` or `protocol.def` — there is no performance series on this branch, `ls -d patches/*/` returns only `pipeasio/`. Acting on the sentence means adding a second bump, and a client at 958 against a server at 957 fails the handshake. Worth fixing before anyone next edits 0078. Fix: drop the clause, say 0078 carries the bump.

- **BASE.txt omits 0079's second gate.** `patches/BASE.txt:283-285` describes 0079 by its private-data marker alone and says unmarked surfaces "are unaffected". The gate has a second clause, `&& !(dirty_rect && IsRectEmpty(dirty_rect))`. For a marked surface `!dxgi_surface_owner_presents(iface)` is false and short-circuits before `IsRectEmpty` is reached, so the empty-rect clause can only ever affect unmarked surfaces — exactly the set the bullet calls unaffected. The patch header discloses it; BASE.txt does not. Fix: add the clause to the bullet.

**Held**

`SERIES.sha256` regenerated to `72630dc7…` and matches the patch file byte for byte; the hunk header went `69 → 72` and patch-check is clean, with the three "no blank line" notes pre-existing on 0008, 0009 and 0054. Build audit passed 103 checks, every patch verified, with 0080 `PASS sha✓ stamp✓` and `pointer_count %u` present in the shipped `ninput.dll`. `probe_stubs.c` re-run on this revision shows no regression — E_HANDLE / E_POINTER / E_INVALIDARG as documented, process survives. The merge against the declared base `main` is clean; main moved six commits (linkd and Link setup, no patches), so the BASE.txt counts are unaffected.

BASE.txt's 0080 bullet not naming `include/interactioncontext.h` is not a defect — the file's entries describe behaviour, not file lists, and 0075 and 0076 are written the same way.

Issue #64, opened 2026-07-25
feature request: touchpad pinching for zooming


hameedoid:
ive tried ableton linux and it works fairly well except for the fact that i can only use ctrl+scroll to zoom

i think it would be nice if the normal trackpad shortcuts were added like pinching to zoom

thanks in advance ^_^

---
[Discord thread](https://discord.com/channels/1307779762189504602/1530636300619747593/1530636300619747593)

shibco:
Implemented about to roll out.

Issue #50, opened 2026-07-21
feature request: click & hold middle-mouse button for arrangement zooming & panning


yioannides:
In Bitwig Studio, holding the middle-mouse button and moving the mouse up, down, left and right allows the user to zoom or pan within and across the arrangement view, like so:

https://github.com/user-attachments/assets/ad521433-f587-4e0e-8ad9-eaa30024e1cd

I vaguely remember a community project called [Live Enhancement Suite](https://enhancementsuite.me/) that essentially bundled [AHK](https://www.autohotkey.com/) scripts, so it would be awesome if we could add this type of QoL feature to the middle-mouse button, cause it's pretty useless now, as-is.

Thank you!

---
[Discord thread](https://discord.com/channels/1307779762189504602/1530636124614295754/1530636124614295754)

shibco:
Also pinch to zoom gestures on trackpads

shibco:
Implemented and about to roll otu!

PR #157, opened 2026-08-08
Middle mouse 2D pan, plus macOS-level smooth mouse zoom gestures

shibco:
The title says it all
~~Fixes~~ _overdelivers_ #64 and #50


patches/0070-winex11-navigate-with-a-middle-button-drag.patch
```diff
@@ -0,0 +1,256 @@
...
++    if (!(event->state & Button2Mask))  /* the release went somewhere else */
++    {
++        middle_drag.active = FALSE;
++        return FALSE;
++    }
++
++    middle_drag.accum_x += pt.x - middle_drag.last.x;
++    middle_drag.accum_y += pt.y - middle_drag.last.y;
++    middle_drag.last = pt;
++    if (abs( middle_drag.accum_x ) > MIDDLE_DRAG_SLOP || abs( middle_drag.accum_y ) > MIDDLE_DRAG_SLOP)
```
ClickSentinel: `middle_drag_delta()` consumes the accumulator to exactly zero on every call — `WHEEL_DELTA / MIDDLE_DRAG_STEP` is 120/24 = 5, an exact integer, so no residual survives. This test therefore measures the per-motion-event step rather than distance from the press, and a drag whose every step stays under the slop scrolls *and* still replays `WM_MBUTTONDOWN`/`WM_MBUTTONUP` on release. Reproducible where motion steps are uniform: a 200 px drag in 1 px steps gives `wheel=71 (sum 1000)` **and** `mbuttondown=1`, against `mbuttondown=0` for the same distance in 8 px steps. Not reachable with a real mouse — one jitter step over 3 px anywhere in the drag latches `moved` for the rest of the gesture — so this is wrong in principle rather than reaching users. `middle_drag.origin` is already stored in the same coordinates. Verified: spurious click gone, plain middle click still reported, wheel totals byte-identical, audit 94/94, and indistinguishable from this branch in a blinded six-trial A/B in Live.

```suggestion
+    if (abs( pt.x - middle_drag.origin.x ) > MIDDLE_DRAG_SLOP || abs( pt.y - middle_drag.origin.y ) > MIDDLE_DRAG_SLOP)
```

patches/0070-winex11-navigate-with-a-middle-button-drag.patch
```diff
@@ -0,0 +1,256 @@
...
++        input.ki.dwFlags = KEYEVENTF_KEYUP;
++        NtUserSendHardwareInput( hwnd, 0, &input, 0 );
++    }
++}
++
++/***********************************************************************
++ *           middle_drag_enabled
++ *
++ * WINE_X11_MIDDLE_DRAG=navigate turns a middle-button drag into navigation
++ * input: vertical movement becomes Ctrl+wheel notches, horizontal movement
```
ClickSentinel: The code passes `control = FALSE` for both axes, and the notes record vertical zoom being dropped on 2026-08-08. `middle_drag_motion()`'s own comment block already describes the current behaviour, so the file contradicts itself — and this is the comment a reader hits first when asking what the variable does.

```suggestion
+ * input: vertical movement becomes wheel notches, horizontal movement
```

patches/0070-winex11-navigate-with-a-middle-button-drag.patch
shibco: BULLISH ON PERF GAINS

patches/0071-winex11-zoom-on-a-touchpad-pinch.patch
```diff
@@ -0,0 +1,220 @@
...
++ */
++static BOOL pinch_zoom_enabled(void)
++{
++    static int enabled = -1;
++    const char *value;
++
++    if (enabled == -1)
++    {
++        value = getenv( "WINE_X11_PINCH_ZOOM" );
++        enabled = !value || (strcmp( value, "off" ) && strcmp( value, "0" ));
```
ClickSentinel: This is a deny-list, so only exactly `off` and `0` disable the feature and every near miss silently enables it. Measured against the built runtime, reading the negotiated XInput2 version as the observable: `off` and `0` give 2.2, but `OFF`, `"off "` with a trailing space, `false`, `no` and `disabled` all give 2.4. Same shape as PR 121's `'off '` parser, and the case that bites is someone disabling the feature to A/B a bug report and getting a false negative. `middle_drag_enabled()` next door is already an allow-list and fails safe; this makes the two symmetric. Verified not to disable `on`, `1`, `notch` or the unset default.

```suggestion
+        enabled = !value || !strcmp( value, "on" ) || !strcmp( value, "1" ) || !strcmp( value, "notch" );
```

patches/SERIES.sha256
```diff
@@ -62,5 +62,7 @@ a07f7ecb2c3a568d8eeae9a42df558cea4d8e279e5bdeb0af67a6649086973f7  0062-winex11-k
 81e02eeac88e6eda5da8044cd5bf74a197b2742f77c46da54e61dbf300031bd3  0064-shell32-route-explorer-folder-open-commands-to-the-h.patch
 7931be1771a070325eb0b87f970e51f7089c216f1bad128fb5ecf504e100e005  0065-win32u-normalize-selected-fullscreen-window.patch
 4b4ca42a11539df9482e7160141be08c10e8f4b1d1c46e1ce3fa60caa0fb993a  0069-win32u-keep-selected-captioned-monitor-sized-window-resizable.patch
+dafcad12cfad452aeb9522194d2e8f07f21f154b063b250bca5c9b387832218f  0070-winex11-navigate-with-a-middle-button-drag.patch
```
ClickSentinel: The suggestions on 0070 change the patch file, so the frozen manifest has to move with it or `build-audit.sh` fails its sha check. This is the hash from a build with all three code edits applied — if it matches after you apply them, they landed exactly as written. If you take only some, skip this and run `./scripts/build-audit.sh --freeze` instead.

```suggestion
2dbda4b400eb40f607bdc6c83329ccc20919119a7c5aeaae0104c228eaa8378b  0070-winex11-navigate-with-a-middle-button-drag.patch
```

patches/SERIES.sha256
```diff
@@ -62,5 +62,7 @@ a07f7ecb2c3a568d8eeae9a42df558cea4d8e279e5bdeb0af67a6649086973f7  0062-winex11-k
...
+2651a6eb829eaf209b5453b15c315e2cb642f3545bd21dee792ad23969cdaae2  0071-winex11-zoom-on-a-touchpad-pinch.patch
```
ClickSentinel: Same for 0071, from the same build. Both manifest lines need to move together with the code suggestions.

```suggestion
15cd86b6e86dc818b4d9d28fbcf7f92441268172accca66a00ba24689ae74201  0071-winex11-zoom-on-a-touchpad-pinch.patch
```

jackson-57:
A great addition to this PR would be to make two finger panning as smooth as macOS.

shibco:
lol, i am a code master

shibco:
the user wantsa clicka da mousea :pinched_fingers:

shibco:
Fixed the serious issues discovered by the long suffering discord members. Sorry!

Also, you can now set inertia scrolling to whatever your little nerd tastes fancy:

eg:
- `WINE_X11_SCROLL_INERTIA_FRICTION=4.5 ableton-live` stops quicker, =1.5 glides longer             
- `WINE_X11_SCROLL_INERTIA_CURVE=linear` for a constant-brake feel
- `WINE_X11_SCROLL_INERTIA=off` if you're more of a stiff kinda freak

shibco:
I am actually going to redo this from scratch

shibco:
i fucka uppa da users mousa :pinched_fingers:

PR #169, opened 2026-08-10
smooth criminal

shibco:
xinput based
pinch to zoom
+ 2d inertia scrolling (via trackpad and mmb)

unlike my dumbass weekend version this one is more agnostic and less ignorantly written.


stickyfran:
i love your naming scheme

shibco:
We are still seeing problems with Interactions and scrolling, despite the fact that this is a fundamental rewrite of the input project. I am begining to suspect these issues exist lower in the wine stack

shibco:
looks like i was right, lmao

ClickSentinel:
Re-reviewed at `29480a0`. Patch-series and runtime level, Live not in the loop: both revisions built and kept side by side (`wine-test-169-022316d6` at 88 manifest entries, `wine-test-169-29480a0` at 91), every harness from the first pass re-run as a two-column arm. `upstream/main` is still exactly this PR's base. 0093 and 0094 were read only — this box has no relative valuators under XWayland, so ClipCursor and raw-motion paths are inert here and their instrumentation never fires.

`patch-check.sh` clean (91 manifest entries, 89 applied), build audit 117/117, registration complete for 0092/0093/0094.

## Blocking

- **An ordinary mouse wheel is dead whenever any button is held, and no setting turns it back on.** Five plain core wheel notches with the left button down, stock behaviour being 5 messages / 600 units: `022316d6` gives sum 600 on every arm; `29480a0` gives sum 0 at defaults, with `SmoothScrolling=disabled`, with all four features disabled, and with **all seven** pointer settings at their most-disabling values — while the no-button control reads sum 600 on both builds, so the probe is not blind. The suppression sits outside every gate, which is the one thing this series' patches are meant never to do, and it reverses 0090's own argument at the previous head, which cited a measurement: "sub-notch wheel with `MK_LBUTTON` set steers the control under the drag (reproduced in Live faders, 2026-08-10)". 0090's prose was rewritten to match, so the reversal is deliberate — but the Live case that motivated the original is now untested. Fix: put it behind a setting, and re-check the fader case before shipping it on.

## Should fix

- **The off switch still takes exactly one spelling, and its warning is invisible.** Re-measured at this head against the negotiated XInput2 version: `disabled` -> 2.2, while `off`, `0`, `OFF` and unset all -> 2.4 enabled, because a miss keeps a default that is on for four of the seven settings. The `unrecognized ... keeping ...` line is a plain `WARN(` on channel `x11drv` and does not survive the launcher's own `WINEDEBUG=-all,+winediag`; the sibling `warn_legacy_env_names` uses `WARN_(winediag)` and does, which is the control that makes that null a fact. Every other switch in the repo is `=off`. Fix: accept `off`/`0` as aliases and move that `WARN` to `winediag`.

## Comment only

- **The coast is now exactly one wheel notch.** Sweeping the drag length across the probe's client edge, `022316d6` emitted 735 units in 116 messages and delivered 0 / 0 / 735 / 735; `29480a0` emits 120 in 16 and delivers 120 on all four. The bounds match `BASE.txt` exactly, and the cap binds rather than the velocity — every arm hits 120 despite different drag lengths. `WHEEL_DELTA` is 120, so a fast throw coasts one notch where it previously coasted about six. Judgement call, not a defect.
- **`BASE.txt:304` still says "0078 carries the `SERVER_PROTOCOL_VERSION` bump"** though 0092 now carries a second and larger one, and 0092's own entry mentions neither the bump nor the `union message_data` growth. Same shape as the note corrected during the PR 156 review. Fix: reword both.
- **All five regenerated patches now trip `patch-check`'s formatting check** ("no blank line before the first `diff --git`" on 0072/0073/0074/0090/0091); before, only the three pre-existing offenders did. Cosmetic, but it is new noise in a check that is otherwise quiet.
- **0093 lands PR 132's fix under a new number while PR 132 is still open.** Not wrong on its own — the port is a rework, not a copy — but two open PRs now carry the same issue-122 fix, and whichever merges second conflicts.

## Held up

- 0094 defaults to `POINTER_WARP_DISABLED` and documents why ("stays disabled pending hardware validation"), which is the inert-unless-selected convention followed correctly. Its near-miss behaviour is safe too, since its default *is* the disabling value — the opposite of the other four.
- 0092's protocol bump is genuinely required, not defensive: it adds a server request (`request_handlers.h`, `request_trace.h`, `user.h`, `winstation.c`) and grows `union message_data`, and `tools/make_requests` auto-bumps on any change to the generated header, so a regenerated tree would have produced a bump on its own.
- `BASE.txt`'s count arithmetic checks: 94 numbers less 0027, 0044 and 0066-0068 = 89 files, matching the applied series and the regenerated `wine-patches: 89`.
- Registration for the three new patches is complete in all four places, including FINGERPRINTS keyed to real strings (`lost X focus to another client while clipping`, `XWayland warp emulation activated after observed failed warps`) rather than stamp-only entries.
- The residual `dist/BUILD-INFO*.txt` diff after a build is `patch-head`, `libusb-pe` and `pipeasio-unix` only — build nondeterminism, consistent with the KB's own rule that two builds of one tree differ; the counts and `patch-stack` agree.
- The 0091 fingerprint moved from `fling start velocity` to `coast start velocity` alongside the trace rename, so the audit is checking a string that still exists.

Everything above is the core X11 pointer path driven by XTEST on Xvfb. The XI2 scroll-valuator path — what a real touchpad and this box's own XWayland pointer actually use — is not reachable from XTEST and was not exercised.

ClickSentinel (approved):

PR #170, opened 2026-08-10
PIPEASIO 1.5, TAKE II

shibco:
@ClickSentinel please test this for CPU usage and stability etc and please test the RT scheduler _vielen dank!_


shibco:
after some extensive testing, this still needs additional work.

shibco:
whoops, broke ci

shibco:
also removed a bunch of fucking slop docs generated by Qwen testing ffs

ClickSentinel:
Reviewed at `4633e9e`. Driver level only — Live was not in the loop, so no CPU or stability figure under a real session. Full container build at head, then 0004, 0006, 0008 and 0009 exercised on the built runtime through a headless ASIO host; the installer's PipeWire gate driven at shell level against the live daemon; the PR's own test suite run locally. 0005 and 0011 are source-read, not run. PipeWire 1.5.85 client and daemon.

## Blocking

### Failed rollback leaves both configs at the other generation

- `scripts/test-pipeasio-installer.sh` fails at head, in CI and locally, deterministically: 40 checks pass, then `not ok - failed rollback did not restore active configuration`.
- After a forced post-swap registration failure the runtime is restored, neither config is: `ableton-wine/config` keeps `live_major=12` instead of `11`, `pipeasio/config.ini` keeps `buffer_size = 512` instead of `883`.
- `rollback.sh` reports the partial failure rather than hiding it, but leaves the original runtime active with both configs describing the generation rolled back to. A retry from that state configures the wrong runtime.
- `rollback.sh` is new here — not in `main`, not in PR 182.

Fix: extend the failure path's restoration to the two config files it already snapshots, or narrow the test's contract if leaving them is deliberate.

## Should fix

### `PIPEASIO_REALTIME` accepts only `on` and `off`

- Measured, one run per value: `on` → `rt thread acquired SCHED_FIFO 15 (was SCHED_OTHER 0)`; `off` → `rt thread stays SCHED_OTHER`; `1`, `true`, `yes`, `enabled` → `WARN: ... is neither on nor off; keeping off`.
- `pipeasio_env_on_off` (0005) is `strcasecmp` against `on`/`off` only. `pipeasio_rt_env_override` for `PIPEASIO_RT_PRIORITY`, in the same header, accepts `0`, `1`, `none`, `default` — the two spellings disagree inside one file, and every other env knob in the repo enables with `1`.
- Not blocking: the warning is emitted without `PIPEASIO_DEBUG`. But under the launcher it lands in `live.log`, not in front of whoever runs the scheduling A/B this PR asks for.

Fix: accept `1/true/yes/enabled` and `0/false/no/disabled`, keep the warning for genuinely unparseable values.

### The new TSan gate fails the build on high-entropy ASLR hosts

- Fails 0/6 here and takes the build with it at step 8 of 8: `FATAL: ThreadSanitizer: unexpected memory mapping`.
- Not the PR's code — a two-line C file built `-fsanitize=thread` fails identically and passes under `setarch -R`. Other stages clean: no-Qt 10/10, ASan+UBSan 11/11.
- GitHub's runner is unaffected, so CI will never surface it.
- `setarch -R` is unavailable inside the container (`failed to set personality to (null): Function not implemented`), leaving a host-wide `vm.mmap_rnd_bits` change as the only way past a mandatory gate — after a full build is already spent.

Fix: make the stage skippable with a logged skip, or run it under a seccomp profile that permits `personality`.

### 0008's pre-1.2 envelope is unreachable through the installer

- 0008 adds a restricted power-of-two envelope for daemons older than 1.2, including a `PAU_DAEMON_RESTRICTED` WoW64 ABI field and its layout guard, on the grounds that Ubuntu 24.04 and Mint 22.x ship 1.0.5.
- `ABLETON_PIPEWIRE_FLOOR` refuses to install below 1.4.2, checked against both client library and daemon. Those daemons cannot reach the envelope through a supported install.

Fix: say which is the intended boundary, drop or document the other.

## Comment only

### RT scheduling measurement

- 2 in / 2 out, 128 frames, 20 s runs: one activation-time xrun with realtime off in two of three trials, none with it on.
- CPU `0.06u / 0.04–0.06s` over ~20 s either way (~0.5%), maxrss ~26 MB, sample position within 900 samples of nominal at 20 s on both arms.
- Output-only, no DSP load — does not answer the CPU-under-Live question the body asks.

## Survived scrutiny

- **Non-power-of-two buffer sizes work end to end (0004).** `pref=300` and `pref=1000` advertised to the host, `CreateBuffers` ASE_OK at exactly those sizes; `31` and `9000` refused with `outside [32, 8192] or unreadable; keeping 256`. Holds through the ASIO API, not just the GUI.
- **Capture resolves and reports (0006).** `PIPEASIO_NUMBER_INPUTS=2` → `channels in=2 out=2`, `CreateBuffers(4 ch, 256 frames)=0`, capture-chain latency line fires. The shipped host-side config seeds `inputs = 0`, so a default run reaches none of it — pre-existing, host-side.
- **Full ASIO lifecycle clean on every arm.** `Init=1`, `CreateBuffers=0`, `Start=0`, `sampleRateChanged 48000`, clean exit.
- **The installer's PipeWire gate is correct.** `ableton_pipewire_version_core` / `_ge` hold across 14 cases including `1.10.0`, `2.0.0`, `1.4.2-1ubuntu1`, malformed input — no string-compare ordering bug. Native probe reports `client=1.5.85 daemon=1.5.85`; preflight passes against the live daemon.
- **The suite's floor test is well built.** It presets `ABLETON_PIPEWIRE_FLOOR=0.3.56` before sourcing precisely to assert the library overrides it.
- **Header/library skew is intentional and documented** at `container-build.sh:154-156` and `Containerfile:117` — SDK link-time only, resolves by soname, no rpath.
- **Shell layer clean.** `bash -n` passes on every changed script; `shellcheck -S warning` clean apart from SC1090 noise.

## Interaction with other open PRs

- **PR 118** carries five pipeasio patches: `0004`, `0005`, `0006`, which this PR rewrites, and `0003` and `0007`, which this PR retires as documented gaps. Whichever lands second reintroduces two numbers the other declared absent, and the audit's gap check cannot see that.

Working on performance metrics now.

shibco:
Ok this is a big PR now with a lot of revisions as it is the "last" one to merge after the new installer. 
This I believe will be the final major release before we migrate to @ClickSentinel's new design.

ClickSentinel:
## Fixed

- **Failed rollback now restores both configs** (`77a2ce2`). Same suite, same machine: **86/86 pass, 0 fail**, where `4633e9e` gave 40 pass and `not ok - failed rollback did not restore active configuration`. The fix is in `lib/manifest.sh` with 36 lines of test beside it.
- **`PIPEASIO_REALTIME` takes `1`.** `pipeasio_env_on_off` → `pipeasio_env_bool` → upstream `pipeasio_parse_bool`, which accepts `1`, `true`, `on`, `yes` case-insensitively and trims whitespace. The two spellings in that header now agree. Source-read; not re-run.
- **The TSan gate no longer fails a whole build on host ASLR — and the fix is tighter than what was asked for.** `lib/tsan.sh` skips only a *recognised* startup incompatibility, only in `auto` mode, and marks the build non-release when it does; races, test failures and unknown errors still exit non-zero. Verified against the first pass's actual failing log, with controls:
  ```
  real log  -> recognised as infrastructure (skipped, non-release)
  race log  -> correctly still fails
  mixed     -> correctly still fails   (race + infrastructure signature in one log)
  ```
  The mixed arm is the one worth having: a genuine race behind an infrastructure signature is still caught, which a blanket skip would have lost.
- **0008 picks one boundary.** Renamed to `enforce-1-4-2-runtime-floor`; `PAU_DAEMON_RESTRICTED` and the pre-1.2 power-of-two envelope are gone, and the driver now enforces the same 1.4.2 floor as the installer — so a post-install PipeWire downgrade or a sideloaded runtime cannot reach an unsupported path either. Source-read.

Looks good to me, with the installer fixes merged in from the previous branch of course. This copy still carries the .run 127 exit code, but fixed in the other branch

ClickSentinel (approved):
approved ensuring the .run 127 exit fixes brought in.

PR #174, opened 2026-08-11
Stop new windows starting black (patch 0091)

shibco:
The Wine base fills every fresh window surface with 0x00, which reads as black on ordinary windows: the floating black squares and the black newly-exposed regions from the graphics review. Patch 0091 fills vanilla white instead and keeps the transparent fill only for 32-bit layered windows, whose dropshadow popups need a transparent first frame. giang17/wine went 0x00 -> 0xff -> 0x00 on this line (53d78989, 44278e2b); their shipped diff is an unconditional 0x00, and this patch limits their per-pixel-alpha rationale to the windows it actually describes.

A/B: `env WINE_SURFACE_INIT=black ableton-live` restores the old fill, `=white` forces vanilla everywhere. The SURFACE-CREATE line of the wmtrace channel (patch 0090) reports the chosen fill per surface.

Verified so far: the patch applies through container-build's git am path with an identical tree, and win32u compiles clean in the build container with no warnings. Runtime verification on the matrix is open: Live dropdowns, WebView2 panes (Learn View, Splice) and plugin dropshadow popups need a look under each fill mode.

Stacked on gfx/trace (patch 0090, the wmtrace channel); this PR targets that branch and merges after it.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,11 @@
 
 ## Unreleased
 
+- New windows and newly exposed areas no longer start black. Plugin
+  popups with transparency keep a clean first frame. If a window now
+  flashes white where it did not before, launch with
+  `env WINE_SURFACE_INIT=black ableton-live` and report it.
+
 - Window problems are now easier to report. Launching with
   `env ABLETON_WM_TRACE=1 ableton-live` logs every window change that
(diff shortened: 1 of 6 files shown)


shibco:
i have killed this because testing has shown it re-introduces issues with Learn View and the Splice integration.

PR #182, opened 2026-08-11
Rewrite the installer to remove technical debt

shibco:
...pre @ClickSentinel's big work upcoming

### DIFF
--- a/BUILDING.md
+++ b/BUILDING.md
@@ -19,24 +19,55 @@ Run:
 
 ```bash
 ./build.sh
-./scripts/install.sh
-./scripts/setup-prefix.sh
-WINEPREFIX="$HOME/.wine-ableton" \
-  "$HOME/.local/opt/wine-d2d1-nspa-11.13/bin/wine" \
-  "/path/to/Ableton Live 12 Suite Installer.exe"
+./scripts/installer.sh install \
+  --live-installer "/path/to/Ableton Live 12 Suite Installer.exe" \
+  --live-major 12 --link=session
 ableton-live
 ```
 
 `build.sh` creates the patched Wine runtime and `dist/ableton-linkd`.
-`install.sh` installs both under your home directory.
+`installer.sh` validates and stages the selected components. If a step fails,
+it restores the state from before the command.
+The lower-level `install.sh`, `setup-prefix.sh`, and `setup-link.sh` scripts are
+component implementations rather than the public command interface.
 
-## Shortcut test tools
+## Manage individual components
 
-The repository includes the tools used to test shortcut support. Run the GNOME
-shortcut test with:
+Use the component commands when you do not want a complete installation:
+
+```bash
+# Copy only the Wine runtime.
+./scripts/installer.sh runtime install
+
+# Create or update only the selected Wine prefix.
+./scripts/installer.sh prefix create --live-major 12
+./scripts/installer.sh prefix update --live-major 12
+
+# Change only Link and the Link state owned by this project.
+./scripts/installer.sh link enable --mode=session
+./scripts/installer.sh link disable
+./scripts/installer.sh link status
+```
+
+Select other locations with `--prefix PATH` and `--runtime-root PATH`. Preview
+all planned file, service, and firewall changes without applying them:
+
+```bash
+./scripts/installer.sh plan update
+./scripts/installer.sh plan uninstall --delete-prefix
+```
+
+The installer does not change system scheduling policy. Run
+`./scripts/setup-realtime.sh` separately if you want to change the host audio
+settings.
+
+## Run the checks
+
+Run the shortcut and installer lifecycle checks:
 
 ```bash
 scripts/test-shortcut-hold.sh
+scripts/test-installer-lifecycle.sh
 ```
 
 Build the Wine menu test with all compiler warnings enabled. Then run its two
@@ -52,10 +83,12 @@ The GNOME test uses temporary data and does not change the desktop settings.
 The Wine test sends keys to its own window. It needs a working Wine display.
 Each mode returns a non-zero status when a required result fails.
 
-Configure Ableton Link networking with:
+## Configure Ableton Link
+
+Configure Ableton Link networking and choose when it runs:
 
 ```bash
-./scripts/setup-link.sh
+./scripts/installer.sh link enable --mode=session
 ```
 
 This requests `sudo` only when an active firewall needs the UDP 20808
@@ -81,12 +114,20 @@ make verify
 
 ## Environment variables
 
+Command-line options override these variables. Exported variables override the
+persistent XDG configuration, and the persistent configuration overrides the
+compatibility defaults.
+
 - `ABLETON_WINE_ROOT` selects the Wine runtime. The default is
   `~/.local/opt/wine-d2d1-nspa-11.13`.
 - `ABLETON_WINEPREFIX` selects the Wine prefix. The default is
   `~/.wine-ableton`.
 - `ABLETON_LIVE_VERSION=11|12` selects a Live major version.
 - `ABLETON_LIVE_EXE` selects one exact Live executable.
+- `ABLETON_LINK_MODE=off|session|always` selects the Link policy shared by the
+  installer, Live launcher, Max launcher, service, and uninstaller.
+- `ABLETON_LINKD` selects the Link daemon path. The generated user unit uses
+  this exact resolved path.
 - `ABLETON_SHORTCUTS=take` temporarily turns off exact Ctrl+Alt+Up and
   Ctrl+Alt+Down entries in the related GNOME settings. Live 11 also turns off
   the exact Ctrl+Alt+Delete entry. The default value, `preserve`, does not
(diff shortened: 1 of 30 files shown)


ClickSentinel:
Reviewed at `c8c02cc4`. Shell level only — no Wine build, Live not in the loop: `bash -n`, `shellcheck`, the lifecycle suite run to completion, a `.run` assembled from this branch's real header and scripts, and both runtime archives extracted and exercised. `upstream/main` is this PR's base commit exactly.

## Blocking

- **The packaged `.run` exits 127 on every successful install.** `setup-run-header.sh:103` delegates with neither `exec` nor `exit`, so bash reads on into `__PAYLOAD_BELOW__` at line 104 and runs it as a command; the old header ended with `exit 0`. Payload 0 → `.run` 127, payload 1/2/42 → 1/2/42 — only success is wrong, since `set -e` exits early on failure. `make-installer.sh`'s `[5/5]` check is `sh "$out" --help`, which returns at line 28 and never reaches the delegation. Fix: `bash "$kit/scripts/installer.sh" "$@"; exit $?` — not `exec`, the EXIT trap still has to remove `$workdir`.

## Should fix

- **The new payload validation accepts the developer `-debug` tree.** `install.sh:157` picks it (`sort -V | tail -n 1` puts `-debug` last), and `validate_runtime_payload()` passes it: the eleven required entries are `bin/wine`, `bin/wineserver` and nine `lib/wine/*`, all of which it has. It has no `bin/wineboot`, `bin/winepath` or `share/wine/wine.inf`, which `setup-prefix.sh:256` and `ableton-live:263,1025` call by absolute path — so install goes green and `wine wineboot -u` then returns `could not exec the wine loader`. Reaches from-source `dist/` trees only; `container-build.sh:234-237` emits one tarball and the kit stages one. Fix: `! -name '*-debug.tar.zst'` in the `find`, and add those three to the required list.
- **`make test` fails on a fresh clone, silently.** The suite needs `./build.sh` to have run — `dist/<runtime>-*.tar.zst` for case 5, `dist/ableton-linkd` for case 19 — which is fair, but `Makefile:20` has no `build` dependency and the abort is a bare `set -e` death: no `not ok`, no message naming the artifact, just 4 `ok` lines and exit 1, which reads as four passing checks. Staged, 4 → 18 → 27/27. Fix: a precondition check that names what is missing.
- **No PR body.** New command surface, new `scripts/lib/`, `setup-run-header.sh` down from 491 lines to 104, one-line commit message. Fix: state the command surface and the compatibility guarantees.

## Comment only

- **The entry-point migration is partial.** `BUILDING.md:31-32` calls `install.sh`/`setup-prefix.sh`/`setup-link.sh` component implementations rather than the public interface, but `Makefile:13` `setup` and `build.sh:49`'s closing `Next:` line still point at them. Fix: point both at `installer.sh`.
- **`ableton-live:252` still writes a fixed `$live_entry.tmp`.** Pre-existing and untouched; noted only because every other shared-state write moved to `mktemp`, and the bring-up `flock` at 219-226 does not cover this block.
- **`installer.sh update --dry-run` with no prefix prints `!! --refresh needs an existing prefix`** (`setup-prefix.sh:144,379`) — a flag that is not in `installer.sh --help`.

## Held up

- All 27 lifecycle assertions pass once the build artifacts are present — every rollback, uninstall, Link-adoption and unit-ownership gate.
- `scripts/lib/` is registered completely, and every path `make-installer.sh` stages exists.
- `bash -n` clean on all 17 changed or added shell files; `shellcheck -S warning` over ~1300 lines of new shell returns two findings, one SC2155 and one SC1090.
- The header's other paths are correct: `--help` and `extract DIR` both exit 0, and a one-byte payload corruption is caught by the sha256 gate.
- The lifecycle harness is properly sandboxed — `mktemp -d` plus `HOME` and every `XDG_*` overridden in `run_isolated`.

ClickSentinel:
The packaged `.run` exits 127 on every successful install.

`scripts/setup-run-header.sh:103` delegates with neither `exec` nor `exit`, so bash returns from the payload installer, reads line 104, and runs `__PAYLOAD_BELOW__` as a command. The header this replaced ended with `exit 0` at line 490.

Only success is affected — `set -e` exits early on any failure, so the fall-through is reached exactly when the install worked. Measured with the real `installer.sh` in the payload, and identical via `sh`, `bash` and direct execution:

| payload installer exits | `.run` exits |
| --- | --- |
| 0 | **127** |
| 1 | 1 |
| 2 | 2 |
| 42 | 42 |

`make-installer.sh`'s `[5/5] wrapper self-check` is `sh "$out" --help`, which returns inside the `--help` case at line 28 and never reaches the delegation, so packaging cannot catch it. README documents `sh …run <command>` as the end-user path at lines 57, 107, 123, 129 and 157.

Fix:

```sh
bash "$kit/scripts/installer.sh" "$@"
exit $?
```

Not `exec` — the EXIT trap still has to remove `$workdir`, which is where the extracted kit lives.

shibco:
This should be ready aaaaa

shibco:
Honestly, fucking terrifying.

ClickSentinel:
Re-reviewed at `ab76e0e2`. Shell level only, Live not in the loop: both revisions kept side by side and every harness from the first pass re-run as a two-column arm, plus `make test` end to end. `upstream/main` is still exactly this PR's base.

Every posted finding is fixed. Measured `c8c02cc` -> `ab76e0e`: `.run` on a successful install 127 -> 0 (1/2/42 still pass through); payload selection with a release and a `-debug` tarball in `dist/` picks `-debug` -> picks the release; a `-debug` tree renamed to a release name goes green -> `!! runtime payload is missing bin/wineboot`, exit 1; the suite on a bare tree gives 4 `ok` and no failure line -> 0 `ok`, both artifacts named, a `not ok` line; `update --dry-run` with no prefix leaks `--refresh` -> names the typed command. `make test` on a staged tree is 42 + 28 = 70 checks, exit 0.

## Should fix

- **Nothing guards the header fix.** `make-installer.sh:127` `[5/5]` is still `sh "$out" --help`, which returns inside the `--help` case at line 28 and structurally cannot reach the delegation at line 105, and the lifecycle suite has zero references to `setup-run-header.sh`, `.run` or the payload marker — so the regression just fixed can return unnoticed. The `--refresh` leak got a new suite case; the blocking one got none. Fix: a suite case that assembles a `.run` from the real header with a stub `installer.sh` and asserts exit 0 on success.

## Comment only

- **`make-installer.sh:21-22` lost its `|| true`.** With `dist/` absent, `find` exits 1 under `set -euo pipefail`, so the script dies before the `!! no …tar.zst in dist/: run ./build.sh first` diagnostic on line 24: no-dist was exit 1 *with* the message, now exit 1 with empty output. Reachable via `make clean` then packaging, and the same silent-`set -e` shape this commit fixes in the suite. Fix: `… -print 2>/dev/null || true)`.
- **`make setup` is no longer idempotent.** `Makefile:13` now runs `installer.sh prefix create`, which refuses an existing prefix (`installer.sh:382-383`, exit 2) where `setup-prefix.sh` created or refreshed; no `make` target exposes the refresh the `##` comment names. Fix: add a target for `prefix update`.
- **`build.sh:49` now points at a command that wants a Live installer.** `./scripts/installer.sh install` reaches `resolve_payload` and noninteractively exits 2 with `!! noninteractive install requires --live-installer FILE or --skip-live-install` (`installer.sh:271-272`, tty-gated, so a human gets a prompt); the line it replaced named `install.sh` and `setup-prefix.sh`, neither of which touched Live. Fix: match `Makefile:10` and append `--skip-live-install`.
- **`ableton-live:252` still writes a fixed `$live_entry.tmp`** — pre-existing and untouched, noted again only because it is now the last shared-state write not using `mktemp`.

## Held up

- The tightened required list rejects nothing real: all three release tarballs here carry `bin/wineboot` and `bin/winepath` as symlinks to a 14400-byte `bin/wine`, so the `-s` test does follow them, plus `share/wine/wine.inf`. The debug tree has `bin/wine` and none of the other three.
- Sourcing `lib/config.sh` at suite level does not escape the sandbox — `run_isolated` uses plain `env`, not `env -i`, but every `export` lives inside `ableton_config_init` (`lib/config.sh:116-118`) and the suite never calls it, so only the unexported `ABLETON_RUNTIME_NAME` is set.
- The new gate's `find … -print -quit | grep -q .` is pipefail-safe: 400 iterations against a populated `dist/`, zero false negatives.
- `bash -n` clean on all five changed files; `shellcheck -S warning` on them returns one SC2155, at `installer.sh:9`, pre-existing.
- `install.sh:159-160` carries the same `find`-with-a-missing-`dist/` exposure as `make-installer.sh` but is byte-identical in behaviour across both revisions, so it is pre-existing rather than this commit's.

shibco:
ya  _k i l l i n g_  me boyo

ClickSentinel:
All four open findings are closed, and the fix went one further than asked: `install.sh:160` picked up the same `|| true`, so a missing `dist/` now reaches `!! no wine-d2d1-nspa-11.13-*.tar.zst payload found` where both earlier heads died mute — the exposure the last pass excused as pre-existing. `make test` is 42 + 29 = 71 checks, exit 0. Nothing new introduced; no findings this pass.

## Held up

- **The new `run-header` case is a real regression test, not a tautology.** Replayed verbatim against each head's header: `c8c02cc` gives exit(0)=**127** and the case fails, `ab76e0e` and `ccabd7b` give 0 and it passes. The third assertion earns its keep too — deleting the single `rm -rf -- "$workdir"` line from the header flips leftover-workdir to 1 and fails the case, so `! find … -name 'ableton-installer.*'` is a live observable rather than a pattern that matches nothing. Worth noting the case invokes the `.run` with no arguments and that is correct: line 28 matches only `--help|-h|help`, line 41 only `extract`, so an empty `$1` falls through to the delegation at 105.
- `make-installer.sh:22` restored: `dist/` absent now prints `!! no …tar.zst in dist/: run ./build.sh first` again at all heads but `ab76e0e`.
- `make refresh` routes correctly — existing prefix passes the gate and stops only on the absent runtime, fresh prefix reports `!! no prefix at …` — and `.PHONY` lists it. `make setup` on an existing prefix still answers in installer vocabulary (`use prefix update`) rather than naming `make refresh`, which is right, since the component cannot know how it was invoked.
- `bash -n` clean on all three changed files; `shellcheck -S warning` on them is clean.
- Suite 29/29 with the artifacts staged, exit 0.

ClickSentinel (approved):

PR #183, opened 2026-08-11
GFX Hardening stack

shibco:


ClickSentinel:
Reviewed at `a89864a`. All 15 patches read against `base-5c23dd1c`, plus registration cross-checks, `patch-check.sh`, a full container build and `build-audit.sh` against the artifact. 0093's gate was measured on the built runtime with a headless probe; every other behavioural claim is source-read, not run. Live was not launched.

## Should fix

### Pending window raise dropped on unlock (0093)

- `window_set_config` records `desired_state.rect` and `desired_state.above` then returns early under `state_locks`; the unlock flush calls it back as `window_set_config( data, data->desired_state.rect, FALSE )`.
- `desired_state.above` is written at `dlls/winex11.drv/window.c:1605` and read nowhere in the tree (checked against `base-5c23dd1c`). `above = TRUE` comes from `sync_window_position`'s top-child case, which `NtUserSetWindowPos` reaches from inside exactly the re-entrancy 0093 defers. The `FALSE` predates the patch — correct for the helper's one old call site in `window_wm_state_notify`; routing every unlock through it is what makes it reachable.
- Given `if (EqualRect(old,new) && (old_above || !above || data->managed)) return;`, it bites on a combined move+raise, where the configure goes out without `CWStackMode`, and on a pure raise of an unmanaged window, where nothing is emitted. Managed top levels with unchanged geometry would not have been raised by base Wine either — the unmanaged case is this stack's own dcomp composition and popup windows.
- The flush also clears the record: `desired_state.above = above` is written above the `if (data->state_locks) return`, so a nested unlock overwrites a pending `TRUE`.
- Reasoned, not measured. A headless probe drove a re-entrant `SetWindowPos(HWND_TOP)` through minimise/restore under the lock, but both arms emitted `mask 0x3 above 0` — the probe window was not the top child, so nothing was pending for the flush to drop. Reaching this needs the locked window to be the top child at `sync_window_position` time.

Fix: pass `data->desired_state.above` instead of `FALSE`.

## Comment only

### `dcomp_teardown_message()` can collide with `WM_NULL` (0103)

If `RegisterWindowMessageW` fails the function returns 0, and the three wndprocs' pre-switch `if (msg == dcomp_teardown_message())` then matches every `WM_NULL` — in `dcomp_swapchain_wndproc` that is an unconditional `DestroyWindow(hwnd)`, and `WM_NULL` is the standard `SendMessageTimeout` liveness probe. Registration failing is remote (atom table exhaustion); the guard is one token.

Fix: `UINT tm = dcomp_teardown_message(); if (tm && msg == tm)`.

### 16-entry cap now sits in the popup/full mode decision (0100)

`dcomp_thread_has_target()` scans `dcomp_popup_stack`, which `dcomp_popup_stack_push` refuses to grow past `DCOMP_POPUP_STACK_MAX` (WARN only). The counter it replaces, `dcomp_subclassed_target_count > 0`, had no cap. Past 16 concurrent targets, a plugin whose *first* target failed to push gets its later targets classified as full windows.

Fix: fall back to the counter when the push was refused, or record that the cliff is accepted.

### Parallel array on an unlocked global (0100)

`dcomp_popup_stack` gains a `dcomp_popup_stack_tid` array kept in sync by two separate `memmove`s in `dcomp_popup_stack_remove`, and 0100 adds cross-thread scans of both. The race is pre-existing; the second array is new and can desync from the first.

Fix: one array of `{hwnd, tid}` so a single `memmove` keeps them consistent.

### 0091, 0093 and 0104 are not class-gated

Unlike the `WINE_X11_FORCE_OFFSCREEN_CLASS` / `WINE_WIN32_FULLSCREEN_CLASS` precedents, these change behaviour for every window in every process under the runtime. Defensible for WM-correctness fixes, and 0093's measurement below is a partial answer — worth saying for the other two.

## Survived scrutiny

- **Registration exact.** 96 wine patches, 96 files, 96 unique numbers; `BASE.txt` says "96 files, numbered 0001 through 0104". `SERIES_GAPS` lists exactly the eight numbers actually missing (0027, 0044, 0066–0068, 0072–0074) — no stale entry of the kind the audit cannot see.
- **The four-digit change is complete.** All five `00*.patch` glob sites fixed (`build-audit.sh` ×2, `container-build.sh` ×3), none left in `scripts`, `.github`, `Containerfile` or `build.sh`, and the numbering loop's `printf '%04d'` already handles 0100+.
- **Build audit 127 checks, every patch verified.** 0090–0104 all fingerprinted in the shipped binary; 0096 and 0099 correctly declared `STAMP_ONLY`.
- **`patch-check.sh` clean.** Manifest 98 entries, hunk counts OK, whole series applies to `base-5c23dd1c`. The three formatting warnings are pre-existing on 0008/0009/0054.
- **All 8 env gates 1:1** between `scripts/ableton-live` and what the patches read, every parser exact-match. No repeat of #121's `'off '`/`-1`.
- **0093's gate measured.** Two overlapping top levels, a re-entrant `SetWindowPos(HWND_TOP)`, minimise then restore, run headless against the built artifact under `+wmtrace,+x11drv`. `WINE_HOST_STATE_LOCK=off` is a genuine off switch — 5 LOCK / 5 UNLOCK unset against 0 / 0 off, balanced, so no lock leak on this path. On and off are request-for-request identical for that lifecycle: same six configure requests, rect for rect, mask for mask, serial for serial. The lock does not perturb an ordinary minimise/restore/move.
- **0093's NULL-sentinel overload is safe across drivers.** It repurposes the existing `pGetWindowStateUpdates` entry point rather than adding one; `nulldrv_GetWindowStateUpdates` (`dlls/win32u/driver.c:878`) returns FALSE ignoring its args, and winex11 is the only other implementation in this tree. The underflow guard is correct, and nested unlocks are absorbed by each setter's own `if (data->state_locks) return`.
- **Base has no `state_locks`** — clean backport target for wine-mirror 852e39a.
- **`git merge upstream/main`** — already up to date.

## Interaction with other open PRs

- **PR 140** touches the same three dxgi files, the same `dcomp_subclassed_target_count` and the same `__wine_dcomp_subclassed` prop. 183's 0094 and 0103 appear to subsume 140's third defect — worth deciding which series owns this code before either merges.
- **PR 132's 0066 applies cleanly on top of this series** (97 patches, verified with `patch-check.sh`), so merge order between 132 and 183 is free.

Issue #165, opened 2026-08-09
Drop and Drop .als and .alc workflow is not suppoirted


0tanh:
## What happened

Click and drag .alc and .als workflow from Ableton Live Timeline to a folder does not automatically create a .alc or .als for use later.  

## Steps to reproduce

1. Create a clip on main live timeline
2. Drag the clip into the folder on the side/Drag the track into the folder
3. "<FolderHierachy>\Untitled.alc could not be opened" for .alc (clips)
4. "<FolderHierachy>\<trackname>.als could not be opened" for .als folders (tracks)

## System summary

Run the profiler on the affected machine from the root of this repository. It
prints a redacted system summary and copies it when a supported clipboard tool
is available.

If the same setup works on macOS or Windows, add that profiler's output for
comparison.

- Linux: `./beta/scripts/ableton-linux-profiler.sh`
- macOS: `./beta/scripts/ableton-macos-profiler.sh`
- Windows: `powershell -ExecutionPolicy Bypass -File beta\scripts\ableton-windows-profiler.ps1`

Review the summary, then paste it here:

```
ableton-linux system summary (Linux)

[SYSTEM]
PRETTY_NAME="NixOS 26.11 (Zokor)"
VERSION_ID="26.11"
Linux 6.18.35 x86_64
CPU(s):                                  16
Model name:                              AMD Ryzen 7 5700X 8-Core Processor
               total        used        free      shared  buff/cache   available
Mem:            15Gi       5.7Gi       2.3Gi        91Mi       7.9Gi       9.8Gi
sys_vendor=ASUS
product_name=System Product Name

[GRAPHICS]
desktop=mango
session=wayland

[AUDIO]
pipewire=active
pipewire-pulse=active
wireplumber=active
pipewire
Compiled with libpipewire 1.6.5
Linked with libpipewire 1.6.5
wireplumber
Compiled with libwireplumber 0.5.14
Linked with libwireplumber 0.5.14
Found "settings" metadata 33
update: id:0 key:'log.level' value:'2' type:''
update: id:0 key:'clock.rate' value:'48000' type:''
update: id:0 key:'clock.allowed-rates' value:'[ 48000 ]' type:''
update: id:0 key:'clock.quantum' value:'1024' type:''
update: id:0 key:'clock.min-quantum' value:'32' type:''
update: id:0 key:'clock.max-quantum' value:'2048' type:''
update: id:0 key:'clock.force-quantum' value:'0' type:''
[...]

Issue #82, opened 2026-07-28
Online Authorization with non-Chromium Browsers does not work


ClickSentinel:
It seems that online authorization only works with Chromium-based browsers - had users with Firefox and Waterfox both unable to authorize online. Switching default browser to Chrome allows online auth to proceed. Needs investigation to track down. 

---
[Discord thread](https://discord.com/channels/1307779762189504602/1531462323762499725/1531462323762499725)

shibco:
We need to diagnose this one pretty badly, because I have no issues at all with Firefox, and neither do any of the testing units that I'm using. This is absolutely confounding, lol

yioannides:
I don't know if there was a regression since the very first release, but I was also able to authorize online via Librewolf at the time...

ClickSentinel:
worth having this issue here since 2 users already reported it, could work with them independently to investigate. `good first issue` tag is gone, but i'd put that on here.

shibco:
this is fixed!

PR #184, opened 2026-08-12
Strengthen .auz and .als support.

shibco:
Addresses #82 by hardening the `ableton:` and `.auz` callback path.

Adds targeted, path-redacted diagnostics for #165; this does not claim to fix or close that issue.


shibco:
This is built from the installer work, so merge #182 first.

ClickSentinel:
- **The per-launch re-pin also overwrites a default the user chose on purpose.** `c8c02cc:ableton-live:167` pinned only after a repair (`[ "$repaired" -eq 1 ] || return 0`); the pin loop now sits outside that guard. Paired arms, real `xdg-mime`, same sandbox, nothing needing repair — `c8c02cc` leaves `application/x-wine-extension-auz=my-own-auz-tool.desktop` in `mimeapps.list`; `3a14aa6` rewrites it to the project ID, with zero repair messages, because the warning at `:207` only fires when the pin *fails* to stick. Fix: put the pin back behind `repaired`, or re-pin only when the current default is one of the project's own retired IDs.
- **A callback that cold-starts Live skips the setup keyed on the Live exe.** `callback_kind` is new here — zero occurrences at `c8c02cc`, where an `ableton://` argument ran full discovery — and `callback_kind != ""` now bypasses it, so `exe_base`/`EXE` stay empty and the `[ -n … ]` gates at `:270` and `:350` fall through — correctly guarded, just not run. With `ABLETON_DPI_MODE=fractional` and a stale menu entry: normal launch → 1 `dpiAwareness` IFEO write and a healed `Name=`; `ableton://` launch → 0 and `STALE NAME`. Moot when Live is already up, which is the usual #82 flow; it reaches a `.auz` opened with Live closed. Fix: hoist the exe-keyed setup above the `callback_kind` branch.
- **Missing `xdg-mime` / `update-desktop-database` / `update-mime-database` takes the runtime install down with it.** `install.sh:221-223` makes them hard requirements and `:249-251` validates everything before any component installs, where `upstream/main` had them optional. Paired arms with those three off PATH: `c8c02cc` exits 0 with `OK: selected components installed transactionally`; `3a14aa6` exits 1 into a rollback, and `--dry-run` can't produce a plan either. Worth caring about because #165's reporter is on NixOS and minimal WM setups often lack `desktop-file-utils`. Fix: scope the requirement to the integration component and skip it with a named warning.

## Comment only

- **`wine-protocol-c74max.desktop` kept the contested name.** Same argument as `ableton:`, and `install.sh:511` pins the default to it. Both that and the unconditional write at `:473-478` (the preserve-foreign test covers only `max9.desktop`) are pre-existing — identical at `c8c02cc:453` and `upstream/main:397` — so this is the missed half of the rename rather than anything new. Worth doing in the same pass.
- **`gtk-update-icon-cache` warns on every install.** `install.sh:513-514` moved to a `!! warning:` while the rollback path at `:85` kept the quiet form, and the local `hicolor` tree has no `index.theme` — that ships with the system package — so a clean successful install always prints `No theme index file`.
- **`test-desktop-integration.sh`'s fake `xdg-mime` echoes back whatever it was set to**, so the new verify-the-pin path can only fail via `ABLETON_TEST_XDG_STICKY=1`. Checked against real xdg-utils across `XDG_CURRENT_DESKTOP` = KDE/GNOME/XFCE/LXQt/unset and it round-trips correctly everywhere, so the logic holds; a real KDE box with `kde-open` takes a different code path and is untested.
- Small ones: `TROUBLESHOOTING.md:268` has a stray `You will` on its own line, and `BUILDING.md:69-70` lists two check scripts after `Makefile:20` added a third.

## Held up

- All three suites pass on the merged tree, including the two #182 added.
- `bash -n` clean everywhere; `shellcheck -S warning` returns SC1090 only, all of it `. "$var"` sourcing.
- The legacy-retirement predicate `^Exec=.*/ableton-live (%u|%f)$` matches what the old templates rendered, and `manifest.sh:62-64,77-79` uses the identical predicate — removal and ownership agree.
- `ableton_remove_managed_file` recording a removed path as owned is right: `manifest.sh:164-168` skips paths that no longer exist, so the stale record retires instead of resurrecting.
- `make-installer.sh:79` stages `desktop/` wholesale, so the renamed templates the new validation requires are in the packaged kit.
- A foreign malformed `mime/packages/*.xml` does *not* abort the install — `update-mime-database` prints `Failed to parse …` and exits 0. Tested because it looked like a second instance of the third open item; it isn't.

The `ableton:` handoff was exercised against a logging fake `wine`, so what's verified is the launcher's argv and ordering, not that the prefix registry resolves the scheme.

PR #194, opened 2026-08-13
tests: stop the release-policy gate breaking on every new patch

ClickSentinel:
`make test` has not completed on `main` since #187. Three independent breakages are stacked, each hiding the next; this fixes all three, plus the reporting gap that made the third one invisible.

**1. The negative series tests named their patches.** `test-release-policy.sh` strips each series' terminal patch and requires `build-audit.sh --check-series-policy` to reject the result. Both names were spelled out, and the Wine one still named `0096-win32u-cache-…` after #188 moved `REQUIRED_WINE_TAIL` to `0097-winex11-restore-pointer-inertia-…`. Stripping 0096 leaves 0097 as the tail, the policy passes, and the test's own `fail` fires. It would break again on the next patch past either tail. Both names now come off `SERIES.sha256` using the same `awk` split `build-audit.sh` uses.

**2. One guard ran under a tool the runner doesn't have.** The check that rejects a release workflow which executes the untrusted installer candidate was `if rg -q …`. `rg` isn't installed on the GitHub runner, and the enclosing `if` swallowed the 127, so the check has never fired there — the log shows `rg: command not found` immediately before a green tick. `grep -E` reads the same pattern.

**3. The runner has no desktop or MIME tooling.** `install.sh` refuses to plan or apply the integration component without `xdg-mime`, `update-desktop-database` and `update-mime-database`; the test-requirements step installed only `cabextract`. Every full-install dry run in `test-installer-lifecycle.sh` failed from the moment that requirement landed in `12af77f` (#170) — only the runtime-only cases, which skip integration, kept passing. #170's own CI died on `.ccache` before reaching `make test`, so it merged red-masked.

**4. That third failure was silent.** Cases expecting a run to succeed send diagnostics to the per-case log and don't check the exit status, so `set -e` ended the suite with no failure line — the CI log jumped from `ok - runtime-only rejects unrelated Link options before mutation` straight to `make: *** Error 1`. The EXIT trap now names the case and replays its log. It reports on a saved copy of the suite's stderr, because the shell can exit while the failing command's redirections are still applied — writing to `&2` there lands inside the very log being replayed.

Verified locally on this branch:

- full `make test`: exit 0, 174 checks, no `not ok`
- `test-installer-lifecycle.sh` with `xdg-mime`, `update-desktop-database` and `update-mime-database` hidden from `PATH`, reproducing the runner: fails with `not ok - a command exited 1 unexpectedly in compat-plan` and the installer's own `!! xdg-mime is required for desktop and MIME integration` above it
- the existing `fail` paths still report once, not twice

Running `make test` needs a built `dist/`; I staged one from another worktree. Two things left alone: the committed `dist/BUILD-INFO*.txt` on `main` records `patches: 104 / wine-patches: 95` against an actual 105/96, and a source-only checkout still can't run `make test` at all.

### DIFF
--- a/.github/workflows/ci-pr-build.yml
+++ b/.github/workflows/ci-pr-build.yml
@@ -120,10 +120,13 @@ jobs:
             [ ! -e "$path" ] || sudo chown -R "$owner" "$path"
           done
 
+      # xdg-utils, desktop-file-utils and shared-mime-info carry xdg-mime,
+      # update-desktop-database and update-mime-database, which install.sh
+      # requires before it will plan or apply the integration component.
       - name: Install installer test requirements
         run: |
           sudo apt-get update
-          sudo apt-get install --yes cabextract
+          sudo apt-get install --yes cabextract xdg-utils desktop-file-utils shared-mime-info
 
       - name: Test installer and launcher lifecycle
         run: make test
--- a/scripts/test-installer-lifecycle.sh
+++ b/scripts/test-installer-lifecycle.sh
@@ -3,8 +3,23 @@ set -euo pipefail
 here="$(cd "$(dirname "$0")" && pwd)"
 root="$(cd "$here/.." && pwd)"
 work="$(mktemp -d "${TMPDIR:-/tmp}/ableton-installer-test.XXXXXX")"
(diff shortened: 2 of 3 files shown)


shibco:
good to go!

PR #203, opened 2026-08-14
QoL fixes for installer v2

shibco:
After the week we just had, can you believe I'm even writing "QoL", @ClickSentinel ?


ClickSentinel:
Like at the end of one's life. We've moved past longevity. We are now focusing on QoL.

shibco:
wow really makes u think

PR #206, opened 2026-08-14
Full installer v2 hardening

shibco:
Combines @ClickSentinel's work along with another, much more awake pass by me.

### DIFF
--- a/BUILDING.md
+++ b/BUILDING.md
@@ -4,10 +4,7 @@ This page covers the current source build, tests, packaging, and maintainer opti
 
 ## Build
 
-Install Podman, Git, Bash, Python 3, GNU Coreutils, GNU Findutils, GNU grep,
-GNU tar, binutils, `zstd`, and `cabextract`. The test suite also needs `ps` and
-`script`. `make check` needs a host C compiler and maths library. Allow about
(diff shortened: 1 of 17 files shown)


ClickSentinel:
Reviewed at `5004c052`. Shell level only — no Wine build, Live not launched. Lifecycle 49/49, PipeASIO 86/86, both exit 0; `shellcheck -S warning` clean on the changed set. Merge with `main` is clean.

Nothing blocking.

- **`mime_clear_default` loses atomicity for something the old form already had.** `cat -- "$tmp" > "$file"` truncates first to keep the inode's permissions, but the `sed -i` it replaces renamed into place — atomic *and* mode-preserving — so a kill mid-write now truncates the user's list. Fix: `chmod --reference="$file" "$tmp" && mv -f`, or note in the comment that surviving a symlinked `mimeapps.list` is the real reason.
- **`disable_link`'s `rmdir` comment overclaims.** "install_unit created this directory" is not necessarily true of `~/.config/systemd/user` — an empty one can predate the install. The `rmdir` itself is safe. Fix: drop the claim.

otherwise, approved, and we close 204 and 203.

shibco:
Fixing those in main. merging now

PR #207, opened 2026-08-14
docs: record the OpenGL EGL/GLX workaround for plugin windows

ClickSentinel:
Wine 11 selects EGL for OpenGL on X11. On some drivers that path does not complete and Wine does not return to GLX once EGL is selected, so plugins that render their interface in OpenGL never get a context and their window does not open. Observed with the NVIDIA proprietary driver, and confirmed fixed by UseEGL=N in the prefix.

The diagnosis greps live.log for the two DPF messages that reach it: both are written through d_stderr and d_stderr2, and the launcher tees only stderr. createNextWindow goes to stdout and would not appear.


shibco:
yes yes good good this makes it into the docs for the next release

PR #216, opened 2026-08-15
Allow updates to replace the managed Live desktop entry

shibco:
Fixes unnecessary blocks during updates

### DIFF
--- a/scripts/test-installer-lifecycle.sh
+++ b/scripts/test-installer-lifecycle.sh
@@ -722,7 +722,67 @@ recorded="$(awk -F '\t' -v p="$live_desktop" '$1=="file" && $2==p { print $3 }' 
 current="$(sha256sum "$live_desktop" | awk '{print $1}')"
 [ "$recorded" = "$current" ] \
     || fail "update did not refresh the Live desktop ownership digest"
+grep -qF "replacing modified managed file $live_desktop" "$base/out" \
+    || fail "desktop entry replacement is not announced"
 ok "update replaces a launcher-healed managed Live desktop entry"
+
+# Uninstall must accept the same launcher-healed entry: the digest mismatch is
+# normal wear, and the entry still routes through the launcher being removed.
+base="$(new_env uninstall-healed-desktop)"
+mkdir -p "$base/fakebin"
+printf '#!/bin/sh\nexit 0\n' > "$base/fakebin/systemctl"
+chmod +x "$base/fakebin/systemctl"
+run_isolated "$base" bash "$here/install.sh" --integration-only \
+    >"$base/first.out" 2>"$base/first.err"
+live_desktop="$base/data/applications/ableton-live.desktop"
+sed -i 's/^Name=.*/Name=Ableton Live 12 Suite/' "$live_desktop"
+printf 'StartupWMClass=ableton live 12 suite.exe\n' >> "$live_desktop"
+run_isolated "$base" env PATH="$base/fakebin:$PATH" bash "$here/uninstall.sh" \
+    --keep-prefix --yes >"$base/out" 2>"$base/err" \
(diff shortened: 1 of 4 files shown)


ClickSentinel:
Head `cd1739c7`, base `58b96561` = `upstream/main`. 3 files, +55/−4.

`ableton_persist_file_prestate` / `ableton_install_file` gain a `collision_policy` parameter defaulting to `protect-modified`; under `replace-modified` a digest mismatch returns 0 instead of refusing. One call site: `install.sh:802`.

Root cause: `ableton-live:279-302` heals the entry on every cold start (`Name=`, `Icon=`, appends `StartupWMClass=`) and never refreshes the manifest. `install.sh` renders it before Live exists — install.sh does not install Live at all — so the heal is unavoidable and the stale digest is normal after a first launch.

Suites on the PR tree: lifecycle **50/50**, desktop-integration 3/3, release-policy 3/3, all exit 0. `bash -n` and `shellcheck -S warning` clean. `git merge --no-ff upstream/main` → already up to date.

## Findings

**1. Uninstall still refuses the healed entry** (should fix, pre-existing). `uninstall.sh:540-572` reads the same digest with no allowance. `harness/uninstall-healed-entry.sh`: control (untouched) → rc 0, `removed …`; healed → **rc 1**, entry left, ownership state retained, `kept modified file …` via `uninstall.sh:876`. Path is install → launch Live once → uninstall.

**2. A user's own edit dies silently** (should fix). Relaxation is by path, not by field. `harness/user-edit-arms.sh` appends `X-UserKeepMe=yes`: base `58b9656` → rc 1, `refusing to overwrite modified managed file`, edit kept; PR `cd1739c` → rc 0, edit gone, nothing printed. A backup is *not* the fix — the early `return 0` at `manifest.sh:968` is what skips the prestate path, and falling through would index the healed entry as pre-install state for uninstall to restore. One echo is the fix.

## Holds

- Converges: `harness/convergence.sh` — after the update the launcher's heal guard (`ableton-live:289-291`) is false on all three fields, digest fresh, uninstall then rc 0.
- New suite case is real: its scenario on the base worktree gives rc 1 and the refusal; it also asserts one `StartupWMClass=`, ahead of `MimeType=`, digest refreshed.
- Other call sites unchanged — both parameters default to `protect-modified`; `cmp -s` at `manifest.sh:961` short-circuits before the policy is read.

scripts/test-installer-lifecycle.sh
```diff
@@ -684,6 +684,106 @@ grep -qF '# user change' "$base/data/ableton-wine/detect-scale.sh" || fail "fail
 grep -q 'refusing to overwrite modified managed file' "$base/err" || fail "modified managed file refusal is explicit"
 ok "update refuses to overwrite a locally modified managed file"
```
ClickSentinel, line 686:
```suggestion

# install.sh installs the version stamp inside an "if !" condition, which
# suppresses set -e, so ableton_install_file must return non-zero for the
# refusal to stop the run.
base="$(new_env modified-version-stamp)"
run_isolated "$base" bash "$here/install.sh" --integration-only >"$base/first.out" 2>"$base/first.err"
printf 'tampered\n' > "$base/data/ableton-wine/VERSION"
if run_isolated "$base" bash "$here/install.sh" --integration-only >"$base/out" 2>"$base/err"; then
    fail "integration overwrites a modified version stamp"
fi
grep -qxF tampered "$base/data/ableton-wine/VERSION" \
    || fail "failed update lost the version stamp modification"
grep -q 'refusing to overwrite modified managed file' "$base/err" \
    || fail "modified version stamp refusal is explicit"
ok "update refuses to overwrite a locally modified version stamp"

```

This was a great fix. We should test for it.

ClickSentinel (approved):

PR #217, opened 2026-08-15
Patch 0099: Fix/jit alloc order

### LINKED ISSUE #214: gen~ devices crash Live on set load: maxclang JIT aborts on IMAGE_REL_AMD64_ADDR32NB
Loading a set containing a `gen~`-based Max for Live device kills Live with an unhandled C++ exception. Max's Clang JIT (`maxclang.mxe64`) fails to link the object it just compiled and reports a fatal error; nothing in the Max or Live stack catches it, so the process dies during document load. The set becomes unopenable — it crashes on every attempt, so the damage is persistent rather than transient.

## Steps to reproduce

1. Create a set and add a `gen~`-based stock device — reproduced here with **Spectral Blur** (two instances inside a rack).
2. Save and close the set.
3. Reopen it.

Live loads the document, reaches the Max device instantiation, and exits. Reproduced 4/4 attempts. The same set with the device removed opens normally.

## The failure

Max's own log records the cause as its last line before the process dies:

```
[2026-08-15 10:20:13.322751 info] [472] Clang compilation error: IMAGE_REL_AMD64_ADDR32NB relocation requires an ordered section layout
```

That string is LLVM's, from `RuntimeDyldCOFFX86_64`. It is a `report_fatal_error` on the path that resolves image-base-relative 32-bit relocations, taken when a section's load address falls below the image base RuntimeDyld derived for the JIT'd object — that is, when the sections were not allocated in ascending address order.

`WINEDEBUG=+seh` confirms the throw and its origin. The exception record:

```
01d8:trace:seh:dispatch_exception code=e06d7363 (EXCEPTION_WINE_CXX_EXCEPTION) flags=1 addr=00006FFFFFC14E53
01d8:trace:seh:dispatch_exception  info[0]=0000000019930520
01d8:trace:seh:dispatch_exception  info[1]=000000000010CD60
01d8:trace:seh:dispatch_exception  info[2]=00000001086B55F0
01d8:trace:seh:dispatch_exception  info[3]=00000001047A0000
```

Resolved against the module list in Max's minidump:

| value | resolves to |
| --- | --- |
| raise address `6FFFFFC14E53` | `kernelbase.dll +0x14e53` (`RaiseException`) |
| `info[2]` ThrowInfo | `maxclang.mxe64 +0x3f155f0` |
| `info[3]` image base | `maxclang.mxe64 +0x0` |

`info[3]` is the throwing module's image base, so the throw is `maxclang`'s own. The SEH handler chain it unwinds through names the caller path, innermost first:

```
maxclang.mxe64 +0x3003ef0   (x3)
MaxLua.dll     +0x47904     (x2)
MaxPlug.dll    +0x14b5754
gen~.mxe64     +0xebd0
MaxAudio.dll   +0x6fb18
Ableton Live 12 Suite.exe +0x24aa2e0  (x6)
Ableton Live 12 Suite.exe +0x24aa2fe  -> _XcptFilter
```

`gen~.mxe64` on that chain ties the abort to gen~ code generation specifically, not to Max device loading in general. Every handler returns 1 (continue search); the exception reaches `_XcptFilter` uncaught and Live terminates.

Correlation across sessions on the same machine, from `Max 9/Logs/MaxPlug*.log`:

| session | clang errors | outcome |
| --- | --- | --- |
| 08:20, 10:06 (set without the device) | 0 | loads fine |
| 10:11:19, 10:12:38, 10:14:08, 10:20:13 (set with the device) | 1 each, always the last line | crash |

Each crash timestamp matches a `Cycling '74/Logs/Max-<timestamp>.dmp` to the second.

## Environment

- ableton-linux runtime dist `2026.08.14.3`, wine-11.13, base giang17/wine d2d1-dcomp-11.13 @ 5c23dd1c, 105 patches
- Ableton Live 12.4.3 Suite, Max 9
- Pop!_OS 24.04 LTS, kernel 7.0.11-76070011-generic
- AMD Navi 31 (RX 7900 XTX), Wayland / COSMIC
- PipeASIO 1.5.0
- 
[Ableton Crash Report 2026-08-15 101222 Live 12.4.3.zip](https://github.com/user-attachments/files/31101615/Ableton.Crash.Report.2026-08-15.101222.Live.12.4.3.zip)
[Ableton Crash Report 2026-08-15 101350 Live 12.4.3.zip](https://github.com/user-attachments/files/31101612/Ableton.Crash.Report.2026-08-15.101350.Live.12.4.3.zip)
[Ableton Crash Report 2026-08-15 101958 Live 12.4.3.zip](https://github.com/user-attachments/files/31101614/Ableton.Crash.Report.2026-08-15.101958.Live.12.4.3.zip)
[ABLETON-WINE-BUILD-INFO.txt](https://github.com/user-attachments/files/31101613/ABLETON-WINE-BUILD-INFO.txt)
[live-crash-seh.log.gz](https://github.com/user-attachments/files/31101616/live-crash-seh.log.gz)
[Live-Log-tail.txt](https://github.com/user-attachments/files/31101620/Live-Log-tail.txt)

[MaxPlug.1.log](https://github.com/user-attachments/files/31101622/MaxPlug.1.log)
[MaxPlug.2.log](https://github.com/user-attachments/files/31101623/MaxPlug.2.log)
[MaxPlug.3.log](https://github.com/user-attachments/files/31101621/MaxPlug.3.log)
[MaxPlug.4.log](https://github.com/user-attachments/files/31101619/MaxPlug.4.log)
[MaxPlug.5.log](https://github.com/user-attachments/files/31101618/MaxPlug.5.log)
[MaxPlug.log](https://github.com/user-attachments/files/31101617/MaxPlug.log)

ClickSentinel:
- Max embeds LLVM 13.0.1. `RuntimeDyldCOFFX86_64::getImageBase()` **memoises** the image base from the first object linked through an instance, then rejects any `IMAGE_REL_AMD64_ADDR32NB` target below it or more than `UINT32_MAX` above.
- The preloader reserves a fixed set of areas. A large application exhausts them, after which anonymous views are placed by the host with no ordering, and the reserved-area path serves only what later frees make room for. A failing set load took **5,203** such fall-throughs against 414 later pool hits; how far a session recovers depends on what it releases.
- The JIT then encodes a relocation across that boundary and aborts: `IMAGE_REL_AMD64_ADDR32NB relocation requires an ordered section layout`.
- Explains the intermittency, and why the device works when added interactively but not when the set is reloaded.

## Observations

Disassembled at the branch reaching `report_fatal_error`, `rax` is `ImageBase` and `rsi` is `Value`. Captured at that branch during a failing load:

| register | value | |
| --- | --- | --- |
| `rax` | `0x24410000` | ImageBase — a slab served from the pool, 608 MB |
| `rsi` | `0x78818fd00050` | Value — a later view placed by unordered mmap, ~132 TB |
| `rcx` | `0x78816b8f0050` | the delta, matching `rsi - rax` exactly |

`rdx` reads `0xffffff01` — `0xffffffff` after `mov $0x1,%dl` — confirming the `> UINT32_MAX` clause, by 132 TB.

## Fix

- When the pool is exhausted, reserve another 256 MB arena and retry; raw mmap stays a last resort.
- Arenas anchor to the **high-water mark of views served from the pool**, so they stay adjacent. Anchoring to a reserved-area end opens a gap (the reserved set includes the main image range); letting host-placed views advance the mark drags it to the builtin region at `0x6ffff...`.
- Growth respects the caller's address range. A request carrying `zero_bits`, or any 32-bit process under WoW64, may only use part of the address space; without the guard each such request that cannot be satisfied still reserves a 256 MB arena outside that range, and 4,000 of them consumed ~1 TB while every one still failed. Stock fails these requests identically, so the guard preserves that behaviour rather than adding a leak on top of it.
- Bottom-up requests stop drawing from the preloader's top area: it serves top-down allocations and ends at `0x7fffffff0000`, Wine's own address-space limit, so nothing can be reserved above it and consuming it strands the pool with nowhere to grow.

## Results

| | before | after |
| --- | --- | --- |
| affected set | crash 4/4 | **opens 3/3** |
| raw-mmap fall-throughs / session | 5,203 | **0** |
| clang errors / session | 1 | **0** |
| alloc mean @ 16k views | 2-3 us | **1 us** |
| alloc p95 @ 16k views | 5-9 us | **2-3 us** |
| alloc worst case @ 16k views | 140-249 us | **6-12 us** |

- Arenas cost address space, not memory: `PROT_NONE | MAP_NORESERVE`, +262 MB VSZ and +36 kB RSS each. 14 per 16k allocations synthetic and 20-21 for a bare `wineboot -u`, every one of them used.

## Testing

- Affected set cold, no priming: 3/3 clean, zero clang errors.
- Set load on an unaffected set, both runtimes on the same NVMe device, zero audio-queue timeouts throughout: 
  - stock median 10.97 s (n=3, range 10.97-11.07)
  - patched median 10.37 s (n=5, range 9.55-12.49). 
  - Ranges overlap; **no difference is resolvable** 
- `jitorder` probe (attached in the thread) — `descents: 2 of 11` on stock, `PASS - monotonic and within a 4 GB window` after.
- `allocbench`, `zerobits` (same attachment) — latency vs view count; ceiling leak 4,053 arenas -> 1.
- Build audit passes, 169 checks.

## Scope

- Both new paths engage **only once `map_reserved_area` has declined outright** — the point at which stock surrenders placement for as long as the pool stays full. Behaviour before that is identical. Gated on `is_win64`.
- Keeps allocations ascending; does not make ordering unconditional. `find_reserved_free_area` still takes the lowest fitting hole, so once anything frees low a later view can be placed below an earlier one — existing behaviour, unchanged. The holes are opened by whatever else is running, not by the JIT, so RuntimeDyld never freeing its slabs does not rule this out. What makes it sufficient is narrower: RuntimeDyld links in a burst, and a burst is when the pool is most exhausted and low holes scarcest.
- Only Live has been exercised with the patch. The stock hazard is not confined to it: @giang17 reproduced `descents: 2 of 11` on Wine 11.0, with fall-throughs during project load in Live 12 Intro (1,026) and FL Studio 2026 (1,140), and none at all in Fender Studio Pro 8.

### DIFF
--- a/patches/0099-ntdll-grow-the-reserved-pool-instead-of-unordered-mmap.patch
+++ b/patches/0099-ntdll-grow-the-reserved-pool-instead-of-unordered-mmap.patch
@@ -1,0 +1,176 @@
+Subject: ntdll: grow the reserved pool instead of falling back to unordered mmap
+
+The preloader reserves a fixed set of areas and a large application exhausts
+them. After that anonymous views are placed by the host with no ordering, and
+the reserved-area path serves only what fits in holes freed since. How much it
+recovers depends on what the application releases: a failing set load fell
+through 5203 times against 414 later hits from the pool. Code that caches an
+image base from its first allocation and encodes later ones as unsigned 32-bit
+offsets from it fails once its allocations land either side of the boundary.
+LLVM's RuntimeDyld does this for IMAGE_REL_AMD64_ADDR32NB, so a Max for Live
+gen~ device aborts a set load when its slabs do.
+
+Reserve further arenas above the highest view served from the pool and retry,
+so views keep ascending and stay on the binary-searched reserved-area path.
+Adjacency matters as much as ordering: an ADDR32NB target must also lie within
+UINT32_MAX of the cached image base, so an arena opened several GB above the
+area a JIT began in fails the same way as one below it. A reserved-area end is
+the wrong anchor, because the reserved set includes the main image range and
+leaves exactly that gap.
+
+Growth respects both ends of the caller's address range. Without the limit_high
+test a request carrying zero_bits, or any 32-bit process under WoW64, reserves
+an arena above the range it may use; without the limit_low test a WoW64 thread
+stack, or an image based above 4 GB, reserves one below it. Either way once per
+failed request: a bare wineboot reserves 76 arenas it never uses, 17 GiB.
+
+The highest preloader area is documented as serving top-down allocations and
+ends at the address space limit, leaving no free space above it, so bottom-up
+requests no longer consume it.
+
+---
+--- a/dlls/ntdll/unix/virtual.c
++++ b/dlls/ntdll/unix/virtual.c
+@@ -2140,6 +2140,111 @@
+  * Try to map some space inside a reserved area.
+  * virtual_mutex must be held by caller.
+  */
++/* The preloader's highest reserved area is documented as serving top-down
++ * allocations, and it ends at the address space limit, so nothing can be
++ * reserved above it. Bottom-up allocations that consume it therefore have
++ * nowhere left to grow and fall through to unordered mmap for as long as the
++ * pool stays full. Keep it for the top-down requests it is meant for, and let
++ * grow_reserved_arena() extend the bottom-up pool instead. The threshold sits
++ * below that area's base at 0x7ffffe000000; nothing is reserved between it and
++ * the area below, which ends at 0x7fff0000, so it selects only that area. */
++#define ARENA_TOPDOWN_BASE  ((void *)0x7ffff0000000)
++#define ARENA_SIZE          (256 * 1024 * 1024)
++
++static void *arena_next;        /* next address to try for a new arena */
++static void *alloc_high_water;  /* end of the highest view served from the pool */
++
++/* Advance the mark a grown arena is anchored to. Both call sites are
++ * reserved-area paths, so a host-placed view never reaches this. The address
++ * test excludes the preloader's top-down area: a view there sits far above the
++ * pool, and anchoring the next arena to it would open exactly the gap this
++ * mark exists to prevent. */
++static inline void note_pool_high_water( void *ptr, size_t size )
++{
++    if (!is_win64) return;
++    if (ptr >= ARENA_TOPDOWN_BASE) return;
++    if ((char *)ptr + size > (char *)alloc_high_water)
++        alloc_high_water = (char *)ptr + size;
++}
++
++/***********************************************************************
++ *           grow_reserved_arena
++ *
++ * Reserve another chunk of address space for bottom-up allocation, above every
++ * chunk reserved so far, so successive views keep ascending once the preloader's
++ * areas are used up. virtual_mutex must be held by caller.
++ */
++static BOOL grow_reserved_arena( size_t size, void *limit_low, void *limit_high )
++{
++    size_t arena_size = max( (size_t)ARENA_SIZE, ROUND_SIZE( 0, size, granularity_mask ) );
++    int attempt;
++
++    if (!is_win64) return FALSE;
++    /* The caller may be restricted to part of the address space - a 32-bit
++     * process under WoW64, or any request carrying zero_bits. An arena outside
++     * that range can never satisfy the request, and reserving one per failed
++     * request consumes address space without effect. */
++    if ((char *)limit_high - (char *)limit_low < arena_size) return FALSE;
++
++    /* Anchor just above the highest view served from the pool. Ascending
++     * placement alone is not sufficient: a 32-bit image-base-relative
++     * relocation also requires its target within UINT32_MAX of the base, so an
++     * arena opened several GB above the area a JIT began in fails the same way
++     * as one below it. A reserved-area end is the wrong anchor, because the
++     * reserved set includes the main image's range, which sits above the pool
++     * and leaves exactly that gap. */
++    if (alloc_high_water > arena_next) arena_next = alloc_high_water;
++    if (!arena_next)
++    {
++        struct reserved_area *area;
++
++        LIST_FOR_EACH_ENTRY( area, &reserved_areas, struct reserved_area, entry )
++        {
++            void *area_end = (char *)area->base + area->size;
++
++            if (area->base >= ARENA_TOPDOWN_BASE) break;
++            /* skip the image's own reserve; it is not part of the pool */
++            if (preload_reserve_end > preload_reserve_start &&
++                area->base <= preload_reserve_start && area_end >= preload_reserve_end)
++                continue;
++            if (area_end > arena_next) arena_next = area_end;
++        }
++        if (!arena_next) return FALSE;
++    }
++
++    /* arena_next is not raised to limit_low. It is static and only increases,
++     * so a single caller with a high lower bound would move it there for every
++     * later allocation, each of which then probes back down the address space.
++     * Return FALSE instead, so such a caller takes the same path as on stock;
++     * the lower bound within an arena is applied by the search in
++     * map_reserved_area. */
++    for (attempt = 0; attempt < 64; attempt++)
++    {
++        void *ptr;
++
++        if (is_beyond_limit( arena_next, arena_size, ARENA_TOPDOWN_BASE )) return FALSE;
++        /* stop before reserving anything the caller could not be given: above
++         * limit_high, or wholly below limit_low. WoW64 thread stacks and
++         * images based above 4 GB pass limit_4g as limit_low, and an arena
++         * under it can never serve them. */
++        if (is_beyond_limit( arena_next, arena_size, limit_high )) return FALSE;
++        if (limit_low && (char *)arena_next + arena_size <= (char *)limit_low) return FALSE;
++        ptr = anon_mmap_tryfixed( arena_next, arena_size, PROT_NONE, MAP_NORESERVE );
++        if (ptr != MAP_FAILED)
++        {
++            TRACE( "reserved new arena %p-%p\n", ptr, (char *)ptr + arena_size );
++            mmap_add_reserved_area( ptr, arena_size );
++            arena_next = (char *)ptr + arena_size;
++            return TRUE;
++        }
++        if (errno != EEXIST) return FALSE;
++        /* something is already mapped there; step over it and try again */
++        arena_next = (char *)arena_next + arena_size;
++    }
++    return FALSE;
++}
++
++
+ static void *map_reserved_area( void *limit_low, void *limit_high, size_t size, int top_down,
+                                 int unix_prot, size_t align_mask )
+ {
+@@ -2170,6 +2275,9 @@
+ 
+             if (start >= limit_high) return NULL;
+             if (end <= limit_low) continue;
++            /* Reserve this area for top-down requests. Serving bottom-up ones
++             * from it leaves the pool with no free address space above it. */
++            if (is_win64 && start >= ARENA_TOPDOWN_BASE) return NULL;
+             if (start < limit_low) start = (void *)ROUND_SIZE( 0, limit_low, host_page_mask );
+             if (end > limit_high) end = ROUND_ADDR( limit_high, host_page_mask );
+             ptr = find_reserved_free_area_outside_preloader( start, end, size, top_down, align_mask );
+@@ -2305,6 +2413,20 @@
+         if ((ptr = map_reserved_area( start, end, host_size, top_down, unix_prot, align_mask )))
+         {
+             TRACE( "got mem in reserved area %p-%p\n", ptr, (char *)ptr + size );
++            note_pool_high_water( ptr, host_size );
++            goto done;
++        }
++
++        /* The preloader's areas are finite and a large application exhausts
++         * them; while they stay full every later view is placed by the host
++         * with no ordering. Extend the pool upward instead and retry,
++         * which keeps successive views ascending and keeps them on the
++         * binary-searched reserved-area path. */
++        if (!top_down && grow_reserved_arena( host_size, start, end ) &&
++            (ptr = map_reserved_area( start, end, host_size, top_down, unix_prot, align_mask )))
++        {
++            TRACE( "got mem in grown arena %p-%p\n", ptr, (char *)ptr + size );
++            note_pool_high_water( ptr, host_size );
+             goto done;
+         }
+ 
--- a/patches/SERIES.sha256
+++ b/patches/SERIES.sha256
@@ -94,6 +94,7 @@ b644c922888e201e1ae525ab22993d89bee835410cd36a739b9f84387009d009  0095-winex11-s
 707a404f2c1d4e02abfa4eb4cbe35669f0caa46ce12dd685d5748f3a699d1186  0096-win32u-cache-the-enumerated-host-font-list-in-the-pr.patch
 80c69a6f555ecc8fcf36fb5d7f466ab9fe4cc1a09f48db7c1cbac442197faef7  0097-winex11-restore-pointer-inertia-and-ignore-held-scroll.patch
 5296ad3f9f91911cf13849e2edf33793d97c2c4d0c06fbee0f119cb06e1114fc  0098-winex11-suspend-XI-scroll-selection-during-core-drags.patch
+84e8c5bf8a7916361449d15abf675815c4a1fa7183f800a23632e00f1e6ae44b  0099-ntdll-grow-the-reserved-pool-instead-of-unordered-mmap.patch
 e2759a5aee2cc5820a3ed27d2d4f954a6a5262c1a95172ad11ffa5302c487387  pipeasio/0001-asio-clamp-unavailable-sample-rate-requests.patch
 b4d53b6ebb24e39c7d23efcbddd837149d4d08886167637a73bb085af851d2c1  pipeasio/0002-asio-report-timeGetTime-in-ASIO-systemTime.patch
 65555e38187564f77ac4f2ba1dd76d812e452c5442e21474b9c8a31706beb7ba  pipeasio/0004-accept-any-buffer-size-in-range-log-every-adjustment.patch
--- a/scripts/build-audit.sh
+++ b/scripts/build-audit.sh
@@ -15,10 +15,14 @@ readonly REQUIRED_PIPEASIO_TAIL='pipeasio/0011-controlpanel-dialog-off-the-host-
 
 check_required_series_tails()
 {
-    local manifest="${1:?series manifest required}" wine_tail pipeasio_tail
+    local manifest="${1:?series manifest required}" body wine_tail pipeasio_tail
     [ -r "$manifest" ] || fail "series manifest is missing or unreadable: $manifest"
-    wine_tail="$(awk '$2 !~ /^pipeasio\// { print $2 }' "$manifest" | sort | tail -1)"
-    pipeasio_tail="$(awk '$2 ~ /^pipeasio\// { print $2 }' "$manifest" | sort | tail -1)"
+    # Read once. Callers may pass a process substitution, and a second open of
+    # that FIFO returns EOF - which reported the PipeASIO tail as absent while
+    # every patch was present.
+    body="$(cat -- "$manifest")"
+    wine_tail="$(printf '%s\n' "$body" | awk '$2 !~ /^pipeasio\// { print $2 }' | sort | tail -1)"
+    pipeasio_tail="$(printf '%s\n' "$body" | awk '$2 ~ /^pipeasio\// { print $2 }' | sort | tail -1)"
     [ "$wine_tail" = "$REQUIRED_WINE_TAIL" ] ||
         fail "Wine series must end at $REQUIRED_WINE_TAIL (found ${wine_tail:-none})"
     [ "$pipeasio_tail" = "$REQUIRED_PIPEASIO_TAIL" ] ||
--- a/scripts/build-audit.sh
+++ b/scripts/build-audit.sh
@@ -10,7 +10,7 @@ SERIES="$root/patches/SERIES.sha256"
 say()  { printf '%s\n' "$*"; }
 fail() { printf '!! %s\n' "$*" >&2; exit 1; }
 
-readonly REQUIRED_WINE_TAIL='0098-winex11-suspend-XI-scroll-selection-during-core-drags.patch'
+readonly REQUIRED_WINE_TAIL='0099-ntdll-grow-the-reserved-pool-instead-of-unordered-mmap.patch'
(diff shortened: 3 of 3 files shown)


ClickSentinel:
@giang17 You may wanna look at this one too. Not sure what other applications do this, but the existing memory addressing scheme Wine uses kind of sucks.

giang17:
Thanks for the pointer — that is a good catch, and the register capture at the
`report_fatal_error` branch makes it hard to argue with.

To answer your "not sure what other applications do this": we have two candidates
installed here, but **no matching symptom so far**, so this is a report of absence
rather than confirmation.

- **Max** — the runtime is present via Live 12 Intro (`Resources/Max`,
  `Max.exe`/`MaxRT.exe`), and 110 `.amxd` files sit in the tree. But **Intro has no
  Max for Live**, so we cannot load any of them. We therefore never reach the JIT
  path at all, and our "no crash" says nothing about your case — treat it as no
  data rather than as a counter-example.
- **Bitwig Studio** — ships LLVM (its licence folder carries the LLVM classpath
  exception), and the Grid is a plausible JIT consumer. That one we could actually
  exercise, but have not.

Nothing in our issue tracker touches JIT, address-space exhaustion or relocation
failures. Our daily testing is Fender Studio Pro 8, FL Studio, Reaper and Live —
heavy on plugin GUIs and Direct2D, essentially free of JIT-compiled devices, which
is the likely reason this never surfaced here.

Two questions, so we can look in the right place rather than guess:

1. **Is the reserved-area exhaustion visible before the crash?** If a session that
   is heading for trouble can be recognised by counting raw-mmap fall-throughs, we
   could check our applications cheaply without needing a failing set. You quote
   5,203 for a failing load — is there a straightforward way to read that number,
   a `WINEDEBUG` channel or a counter, or did it come from instrumentation?
2. **Does it need a large set specifically, or just a long-running process?** Put
   differently: is the trigger total allocation volume, or the burst of a set load?
   A DAW left open for hours with plugin GUIs opening and closing allocates a lot
   over time — if volume is what matters, our workload might reach it by a
   different route.

We patch `ntdll` already (AF_UNIX reparse points, `MADV_FREE` for `MEM_RESET`), so
picking this up would not be new ground for us. Holding off until it settles in
your tree, though — without a reproducing case here we would have nothing to
verify against, and a change this deep in memory placement is not something we
want to carry untested.

We cannot help on the Max side without a Live edition that has Max for Live.
Bitwig we could exercise — if you have reason to think its Grid hits the same
wall, say so and we will try to provoke it here for a second data point on
different hardware.

ClickSentinel:
**1. It is a stock trace, one grep.** Both lines are stock `TRACE`s in `map_view` — `virtual.c:2307` and `:2328` in our 11.13 base and at the same two lines in 11.15, so it reads the same on your tree:

```
WINEDEBUG=+virtual wine ... > /tmp/v.log 2>&1
grep -c 'got mem with anon mmap'   /tmp/v.log   # placed by the host, unordered
grep -c 'got mem in reserved area' /tmp/v.log   # served from the pool
```

The 5,203 is that grep, no instrumentation. Check your launcher does not export `WINEDEBUG=-all` first — ours does, and the count then reads a false zero.

It will not warn you in advance, though. The low area declined its first request 15 allocations before the first fall-through: `0x110000-0x68000000` (1,662 MiB) fills, then the 16 MB window at `0x7f000000` yields its last 15 × 64 KB, then the top-down area at `0x7ffffe000000` is already full. A non-zero count means the session has crossed, not that it is heading there.

Cheaper, no tracing: pool views sit below `0x68000000`, host-placed ones land in the mmap region (`0x7744…` in our capture). Anonymous entries above `0x700000000000` in `/proc/PID/maps` run to thousands once a process has crossed, single digits when it has not — coarse, since the kernel coalesces same-permission neighbours, but free.

**2. Neither — peak concurrent footprint in a fixed window.** The pool is preloader address space (`preload_info`, byte-identical in 11.13 and 11.15), not a counter that climbs, and unmapping returns space to it. So lifetime volume does not accumulate toward the wall: hours of plugin GUIs opening and closing stay clear as long as the space goes back. What crosses it is peak simultaneous footprint — a set load instantiating hundreds of devices at once, ~10,270 pool-served views in our case. The qualifier is anything retained, plus fragmentation: 2,426 of the 5,203 fall-throughs were 64 KB requests, so the window had no 64 KB hole left. A long session with a growing resident set could reach it that way, but we have not run one to exhaustion — mechanism, not measurement.

For your four, the first question is not whether they JIT but whether they ever get an anonymous view above `0x100000000000`. If they never do, the ordering never matters.

**Testing it without Max for Live.** The hazard is in `ntdll`, so you need neither Max nor a JIT. The probe below blocks the address `SectionMemoryManager` hints at, forcing the no-hint retry, and reports where the retry lands relative to the previous slab: `descents: 2 of 11` on stock, `PASS - monotonic and within a 4 GB window` after. Seconds to run, needs only a prefix — running it on your hardware is worth more than hunting for an affected application.

Attached, with the two bench tools behind the latency and `zero_bits` numbers. Plain console PEs, no Wine headers needed: `x86_64-w64-mingw32-gcc -O2 -o jitorder.exe jitorder.c`.

**Bitwig** I would not single out. Shipping LLVM is not the discriminator; a COFF `RuntimeDyld`/MCJIT path emitting `IMAGE_REL_AMD64_ADDR32NB` is, and crossing the window is the precondition. Check the maps count on a heavy Grid patch first — if nothing lands above `0x100000000000` there is nothing to provoke.

Holding off is the right call and I would do the same. `map_view`'s fall-through path is unchanged between 11.13 and 11.15 if you do want to try it, but only Live has been exercised here.
[jit-tools.zip](https://github.com/user-attachments/files/31116555/jit-tools.zip)

giang17:
Ran `jitorder.exe` and both greps here. Thanks for packaging the probe — being able
to test the hazard without Max or a JIT is what made this cheap enough to actually do.

**Setup:** `giang17/wine` `d2d1-dcomp-11.0`, so Wine **11.0** rather than your 11.13
base. X11, NVIDIA. Same prefix our audio work runs in.

## The probe fails identically

```
round  9: 0x00007ffffffaf0000   ascended  by 0x200000
round 10: 0x000074be8caf0000   DESCENDED by 0xb4173000000   [BELOW FIRST SLAB]
round 11: 0x000074be8c8e0000   DESCENDED by 0x210000        [BELOW FIRST SLAB]

descents: 2 of 11
verdict: FAIL - a later slab landed below an earlier one
```

**`descents: 2 of 11`, the same number you quote for stock** — so the fall-through
path is unchanged from 11.0 through 11.15, not only between 11.13 and 11.15.

## Three applications, three different answers

`WINEDEBUG=+virtual`, cold start to loaded project, then 30 s of transport:

| | `anon mmap` | `in reserved area` | share | when |
|---|---|---|---|---|
| Fender Studio Pro 8 (no project, idle main window) | **0** | 2 434 | 0 % | — |
| Ableton Live 12 Intro, "Neon Drive" demo set | **1 026** | 15 688 | 6 % | **1 025 of them in one window** |
| FL Studio 2026, "Thief" demo project | **1 140** | 4 757 | **24 %** | **spread across the whole run** |

So we do cross the wall, on two of three. Studio Pro never gets near it — highest
pool-served address `0xf8b0000`.

Your warning about `WINEDEBUG=-all` was exactly right, by the way: our launch scripts
set it everywhere. Run that way both greps read zero, **including the pool counter**,
which is the tell that the measurement is worthless.

## Where our numbers disagree with "permanently"

You write that once the areas are exhausted, every later view is host-placed
"permanently, and the reserved-area fast path is never used again". That holds for
FL Studio here, but **not for Live**:

- **Live**: 1 025 of the 1 026 fall-throughs sit between 46 % and 65 % of the log,
  then it stops — a single straggler appears later, at 99.7 %. Meanwhile the pool
  counter keeps climbing: 12 925 by the end of loading, 15 688 after 30 s of
  playback. So the pool was used again, heavily, after the burst.
- **FL Studio**: fall-throughs continue throughout, 565 of 1 140 in the last third
  of the run. It does not recover.

Both fit your own mechanism — the pool is address space and unmapping returns it, so
whether it recovers depends on what the application retains. It just means the
"permanently" in the description is a property of the workload rather than of the
code path, which might be worth softening in the commit message: someone measuring
Live and seeing the counter recover could conclude the diagnosis does not apply to
them.

## One caveat on the cheap `/proc/PID/maps` indicator

93 anonymous entries above `0x700000000000` in the same Studio Pro session that
produced **zero** fall-throughs. Those come from elsewhere — thread stacks, large
individual allocations.

- A **two-digit** count says nothing. Your "single digits versus thousands" holds at
  the ends, but the middle is not empty, and 93 looks alarming until the trace is
  next to it.
- The count also needs Wine's own top-down area excluded or it drowns in it: a naive
  `>= 0x700000000000` filter gave us 985, of which 39 were the preloader's area at
  `0x7ffff…` and 904 were below the threshold once parsed properly. 93 were actually
  in the host mmap region.

## Where that leaves us

The precondition is met on two of our three DAWs. What is missing is the second half
— a COFF `RuntimeDyld`/MCJIT path emitting `ADDR32NB`, which none of them has as far
as we know. So nothing breaks here today, but the ground is prepared, and in FL
Studio permanently so.

We still hold off on the patch: no reproducing crash means nothing to verify a fix
against. But the calculation changed — this is no longer "probably irrelevant for
us", it is "one missing ingredient away".

Happy to re-run any of this if a particular question would help. The three logs are
kept.

ClickSentinel:
Yeah, you are right. My /proc/maps figure wasn't correct. "Single digits" was a count of trace lines, not maps entries. I never measured maps at all, oops. Your 93 is the real number. Agreed on excluding 0x7ffff… too.

11.0 checks out structurally too: the map_view fall-through block is the same 11.0 to 11.15, and preload_info is also the same across 11.0, 11.13 and 11.15. The 2 of 11 match isn't luck. fixed 32 MB area, fixed 1 MB slabs.

"Permanently" is fixed, in the patch header, both code comments and the commit message. Ours recovered too, barely: 414 pool hits against 5,203 fall-throughs. Recovery isn't safety though — RuntimeDyld only needs the line crossed once while it holds a cached ImageBase.

One ask if you still have the logs: size distribution of FL's fall-throughs. Ours with Max was 2,426 of 5,203 at 64 KB, which is what shows the window had no small hole left rather than no large one.

giang17:
Logs were kept, so here is the size distribution. It splits into two clearly
different modes, and only one of them looks like yours.

| | exactly 64 KB | ≤ 64 KB | ≥ 1 MB | distinct sizes | largest |
|---|---|---|---|---|---|
| your Max set | 2 426 / 5 203 (46.6 %) | — | — | — | — |
| **Live 12, "Neon Drive"** | **845 / 1 026 (82.4 %)** | 84.5 % | 10.2 % | **17** | 16 MB |
| **FL Studio, "Thief"** | **16 / 1 140 (1.4 %)** | 15.2 % | **48.2 %** | **166** | **512 MB** |

**Live is your case, more so than Max.** 845 of 1 026 requests are exactly one
allocation granule, from a repertoire of only 17 sizes. The window had no small
hole left, exactly as you describe.

**FL Studio is a different failure.** Almost nothing at 64 KB; nearly half the
fall-throughs want 1 MB or more, one asks for 512 MB, and the requests come in 166
distinct sizes. That is not "no small hole left" — it is no large *contiguous* run
left, in a pool whose bottom-up area (`0x110000-0x68000000`) is about 1.6 GiB to begin with.

Which also explains the recovery difference we reported:

- **Live** frees and re-requests granule-sized views, so 64 KB holes reappear and
  the pool becomes usable again — hence the burst-then-quiet pattern and 15 688
  pool hits afterwards.
- **FL Studio** needs multi-megabyte contiguous runs. Once the pool is chopped up,
  those do not come back from ordinary unmapping, so it keeps falling through for
  the rest of the session.

So the 64 KB share is a good signal for the fragmentation mode, but it is not the
signal for exhaustion as such — a workload can cross the line with almost no small
requests at all. If your arena growth anchors 256 MB at a time, FL's shape is
probably the friendlier one for it: a fresh arena satisfies a 2 MB or 8 MB request
immediately, whereas a pool that is merely fragmented at granule size may keep
finding *some* hole and stay in the slow path longer.

Worth noting for the "recovery isn't safety" point you make: agreed, and FL shows
the sharper version of it. Its fall-throughs never stop, so any JIT loading at any
point during that session would be exposed, not only one loading during a burst.

Numbers from the same two runs as before; raw traces still here if you want a
different cut of them.

---

Two things from reading the current head, now that you have confirmed `preload_info`
is identical across 11.0, 11.13 and 11.15 — that removes the reason we held these
back, since a remark from our tree applies to yours.

**`ARENA_TOPDOWN_BASE` is 224 MB below the area it names.** The preloader's top-down
area starts at `0x7ffffe000000`; the constant is `0x7ffff0000000`. It works, because
nothing is reserved between the third area (ending `0x7fff0000`) and the fourth, so
the comparison catches exactly the intended one. But it is a magic number that goes
quietly wrong if the layout ever shifts, and `grow_reserved_arena()` already walks
`reserved_areas` for the `!arena_next` case — the highest entry could come from there
instead.

**`note_pool_high_water()` promises more than it filters.** The comment says "Only
views served from the bottom-up pool count", but the code tests the address alone, so
a `top_down` request satisfied below the threshold still advances the mark. Probably
harmless — we could not construct a case where it hurts — but the next reader will
spend time reconciling the two.

A third, less certain: if `grow_reserved_arena()` succeeds and the following
`map_reserved_area()` still fails — a large `align_mask` finding no suitably aligned
run in the fresh 256 MB, say — the arena stays reserved and unused. The `limit_high`
guard covers the `zero_bits` case you describe, not that one. Whether it can happen in
practice we have not established.

None of these is a blocker, and we have not built or run the patch — this is reading,
against 11.0 sources, with the structural equivalence you established.

ClickSentinel:
The size split is a really good answer. FL is clearly a different mode. no large contiguous run left, which the 64 KB share would never have shown.

Remark 3 didn't reproduce: five 300 MB requests at 2 MB alignment all landed inside arenas. mmap_add_reserved_area merges adjacent ones, so the padding gets absorbed by the neighbour.

Looking for it found worse, though. `grow_reserved_arena` tests `limit_high` and never `limit_low`, so a WoW64 thread stack - limit_4g as its floor - reserves 256 MB it can't use, once per thread, until arena_next climbs past 4 GB. A bare wineboot -u: 76 arenas, 74 of them never used, 17.25 GiB. With a floor test: 21 arenas, 21 used, 5 GiB. Latency unchanged, jitorder still PASS, zerobits still 1 arena.

1 and 2 agreed, fixing both. On 1, careful with deriving it from reserved_areas - the list merges and grown arenas join it, so "highest entry" stops meaning the preloader's area.

Thanks. That one was live on every process, not a corner case.

giang17:
Point 3 was rightly dropped. Five 300 MB requests at 2 MB alignment settle it, and
the merge behaviour of `mmap_add_reserved_area` explains why the ordering argument
does not hold. We had flagged that one ourselves as "whether it can happen in
practice we have not established" — it can't, and that is the useful outcome.

Your objection to point 1 is correct too. Since `reserved_areas` merges adjacent
entries, and a grown arena joins that same list, "highest entry" stops meaning
"the preloader area" the moment growth happens. Documenting the invariant in a
comment is the better fix; our suggestion would have encoded an assumption that
the patch itself invalidates.

**On `limit_low`: we measured it here, and your fix holds on a different tree.**

`dd33331cadb` applies to Wine 11.0 with `git am --3way`, no conflict. Measured with
a bare `wineboot -u` into a fresh prefix, `WINEDEBUG=+virtual`, counting `map_view`
outcomes:

| `map_view` outcome | stock 11.0 | with your patch |
|---|---|---|
| reserved area | 1063 | 987 |
| **unordered `anon mmap`** | **17** | **0** |
| `map_free_area` | 10 | 83 |
| grown arena | 0 | 20 |
| **total** | **1090** | **1090** |

Two things worth reporting:

**The fall-throughs go to zero, and the total is unchanged.** 1090 either way — the
patch redirects allocations, it does not add any. That is the property one wants to
be able to state about a change in `map_view`, and it is nice to have it as a number
rather than as a reading of the code.

**No arena waste on this tree.** 20 arenas reserved, 20 used — one 256 MiB arena per
process across 20 processes, and not a single process that reserved without ever
being served from it. So whatever produced your 76-with-74-unused, the `limit_low`
test removes it here as well; there is nothing left to trim on 11.0.

Note this is 11.0, not 11.15, so it is a second tree rather than a second run — which
is what makes it worth posting. The structural agreement you found (identical
`map_view` fall-through block, identical `preload_info` across 11.0/11.13/11.15) now
has a behavioural counterpart.

We did not reproduce the pre-fix state, so we cannot say whether 11.0 ever showed the
76/74 shape — only that with `dd33331cadb` in place it does not.

giang17:
Ran FL Studio under your patch on 11.0. **2 020 fall-throughs to zero**, no arena
waste, and the cost lands where you'd want it.

Demo project "Ookay - Thief", cold start to loaded project, then 30 s of transport,
`WINEDEBUG=+virtual`, counting `map_view` outcomes. Same procedure both runs,
transport clock verified at 0:30:00 / 0:30:01:

| `map_view` outcome | stock 11.0 | with `dd33331cadb` |
|---|---|---|
| reserved area | 4 351 | 6 302 |
| **unordered `anon mmap`** | **2 020** | **0** |
| `map_free_area` | 621 | 812 |
| grown arena | 0 | 30 |
| **total** | **6 992** | **7 144** (+2.2 %) |

**The fall-throughs move into the regular reserve almost one-for-one** — `reserved
area` gains 1 951 of the 2 020. Only 30 allocations come out of the growth path
itself: a grown arena joins the reserved list and is then served by the ordinary
binary-searched lookup. The expensive branch is entered once per arena, not once per
allocation. Totals agree to 2.2 %, which is noise across two separate FL launches.

**No waste under load.** 30 arenas across 22 processes, all 22 served from what they
reserved. Sizes adapt — mostly 256 MiB, several at 512 MiB.

FL's shape was the case I was least sure your growth would cover: 174 distinct
sizes, **58.4 % at ≥ 1 MB**, median 2 MB, only 3.5 % at exactly 64 KB. That is
fragmentation of large contiguous runs, not exhaustion of small holes — and it is
covered.

### One thing your arenas cannot catch, for completeness

The stock GPU run also contained **two reservations of ~1.4 TB each**
(`0x73260e200000-0x747f0e1f0000` and `0x747f1fff0000-0x75d71fff0000`, same process).
Those are almost certainly the graphics driver reserving address space — they do not
appear at all in headless runs on llvmpipe. At three orders of magnitude above the
largest arena your patch ever created here (512 MiB), such a request must keep
falling through.

They did not recur in the patched run, so this is not an observation of them falling
through — only a note that they cannot be served from an arena, and that they land
exactly in the far region (`0x73…`) we originally reported as "tens of terabytes from
the pool". If a JIT ever caches an ImageBase near one of those, arena growth would
not be what saves it.

Not measured: wall-clock cost. This run counted allocations, it did not time them.

shibco:
hello friends, I just reviewed this myself and ran a couple of tests. 
Looks good, I think this is ready to go.

Accepting and merging. Good work!

PR #212, opened 2026-08-15
Split general fixes from Flake work in PR #27

realitymolder:
This PR extracts the general fixes originally included in PR #27:

The changes are separated from the Flake-specific work so the updated PR can focus only on the Nix/Flake implementation.

The branch was rebased onto the latest upstream state before the split. All 21 general-fix commits were preserved and replayed linearly.

The corresponding Flake work is available in a separate PR:

https://github.com/Version33/ableton-linux/pull/1

Generated Chgangelog:
 
# PR #212 Changelog

This PR contains the general fixes separated from the Flake work in PR #27.
It is based on the synced `shibco/main` at `58b9656145b6437b78e839e845c28b88adb57802`.
No Nix/Flake packaging changes are included intentionally.

## Installer and Setup

- `6ebcc9a` makes Ableton Live auto-install opt-in through `ABLETON_LIVE_AUTOINSTALL=1`.
- `f114e7b` derives the Live major from the selected installer when no version is pinned.
- `e4cbdfa` sorts installer payloads by versioned basename and selects the matching prefix recipe.
- `6ec2de1` kills a stale wineserver before the first prefix settle.
- `d1a06f9` reports when the discovered installer major differs from the prepared recipe.
- `63e8474` prevents the settle poll from outliving a fast wineserver wait.
- `4ececf8` selects installer flags per engine and removes the obsolete audio-driver task.
- `4adda66` selects the newest installer by version instead of edition.
- `f2a877c` bounds the post-wineboot settle operation.
- `0075619` simplifies the cabextract host-tool check.

## Documentation and Configuration

- `81f6632` was authored against the then-current PipeASIO 1.2.2 base and
  documented `config.ini` rather than the proposed `PIPEASIO_*` surface.
- `350447e` immediately reverts that documentation change. These commits are
  preserved as part of the historical general-fix sequence, but they do not
  introduce or document current PipeASIO 1.2.2 behavior.
- The current synced Flake/main uses PipeASIO 1.5.0. This PR does not change
  the PipeASIO implementation, version, patches, or configuration model.
- `7555f00` keeps detect-theme indentation aligned with upstream style.
- `4791f58` declares the shell for the remaining sourced libraries.

## Test and Tooling Hardening

- `f0007b3` keys the shortcut-test GSettings stub on its own arguments.
- `ffa2d52` makes the tester-kit-created prefix path private.
- `5ea87b1` matches patch filenames rather than formatted directory listings.
[...]

PR #167, opened 2026-08-09
fix(launcher): terminate lingering wineserver process when main window closes

stickyfran:
### Problem
When closing the main Ableton Live window (or standalone Max 9), the Wine process terminates, but a residual `wineserver` daemon process frequently remains active in the background for the `$WINEPREFIX`.

### Proposed Fix
While modifying the desktop entry (`Exec=bash -c "~/.local/bin/ableton-live; WINEPREFIX=~/.wine-ableton wineserver -k"`) works as a workaround for desktop menu shortcuts, implementing the cleanup directly within `scripts/ableton-live` and `scripts/max9` provides a complete, robust fix:

1. **Launcher-level Cleanup**: Instead of `exec` replacing the shell process, the launchers now run `wine`, wait for process exit, and invoke `"$WINESERVER" -k 2>/dev/null || true` before exiting with wine's return code.
2. **Universal Coverage**: Works seamlessly whether Ableton Live is launched from the desktop menu (`.desktop`), terminal, custom DAW scripts, or external launchers.
3. **Environment Aware**: Automatically uses the launcher's `$WINEPREFIX` and `$WINESERVER` (`$WINE_ROOT/bin/wineserver`), avoiding hardcoded prefix paths or global binary assumptions.

### Verification
- Tested syntax with `bash -n scripts/ableton-live` and `bash -n scripts/max9`.
- Verified exit code forwarding and cleanup execution flow upon process exit.

AI disclosure: Used ai for translation and debugging but the proposal is mine and already tested on Live.

PR #220, opened 2026-08-16
Fix: Stop the WebView2 updater wedging installs and sessions

### LINKED ISSUE #213: Installer fails on fresh re-install (after uninstalling)
Possibly relevant: #211 

## System

- Ableton Live edition and version: Suite 12.4.3
- Ableton Linux project release: 2026.08.14.3
- Linux distribution and version: Fedora 44
- Desktop and session type, such as GNOME on Wayland: GNOME / Wayland

## Problem

I was having some potentially Fedora-specific issues (re: #72), so I decided to uninstall my existing prefix (via the `uninstall` flags and deleted some additional cache myself) and reinstall Ableton Live with the newest ableton-linux release. The installer fails after installing Live and no prefix is being built (it's worth mentioning I wasn't able to update either, before doing all of the above), with the final terminal output lines being:

```
...
0134:fixme:wtsapi:WTSUnRegisterSessionNotification Stub 0001008C
0134:fixme:uxtheme:BufferedPaintUnInit Stub ()
!! post-installer prefix wait timed out
!! installer transaction failed; restoring the previous component and prefix state
!! rollback complete; failure record: /home/yiannis/.local/state/ableton-wine/transactions/installer.THrclD/FAILURE
```

## Steps to reproduce

1. `./install-ableton-latest.run uninstall --delete-prefix`
2. `./install-ableton-latest.run install --live-installer ableton_live_suite_12.4.3_64.zip --link=session`

## Relevant logs

**FAILURE:**

```
command=install
exit=1
restoration_complete=yes
restoration_error=
rollback_log=/home/yiannis/.local/state/ableton-wine/transactions/installer.THrclD/rollback.log
```

**rollback.log:**

```
-- rollback started for install (exit 1)
-- prefix rollback preflight
-- component rollback preflight
-- Link rollback preflight
-- prefix rollback
-- component rollback
-- Link rollback
```

### LINKED ISSUE #215: Installation : Post-installer prefix wait timed out
Distro : Fedora 44

Terminal error:
! post-installer prefix wait timed out
!! installer transaction failed; restoring the previous component and prefix state
!! rollback complete; failure record: /home/kami/.local/state/ableton-wine/transactions/installer.aBKv3r/FAILURE

FAILURE log: 
command=install 
exit=1
restoration_complete=yes
restoration_error=
rollback_log=/home/kami/.local/state/ableton-wine/transactions/installer.aBKv3r/rollback.log 

Rollback log:
-- rollback started for install (exit 1)
-- prefix rollback preflight
-- component rollback preflight
-- Link rollback preflight
-- prefix rollback
-- component rollback
-- Link rollback

This displayed in my terminal as I was installing Ableton, though after this error - Ableton had installed perfectly and everything is functioning.

Just putting it here if anyone runs into more issues from this certain error ^^

ClickSentinel:
Fixes #213, #215 

Includes and supercedes #167 

Will impact #212 

Live 12's installer runs the WebView2 bootstrapper, which starts `MicrosoftEdgeUpdate.exe /c`. Under Wine it cannot validate its COM registration, cannot start its update worker, and so parks in `Core::DoRun`, no window, 0% CPU, holding the prefix's `wineserver` open indefinitely. Its own log says: `ShouldRunForever: 1`, `Will stay running`. Which is kind of ridiculous honestly.

That wedged process caused two separate failures.

**Installs were destroyed.** `install_live_payload` ends with a 60s `wineserver -w` and treated any non-zero exit as fatal. The updater never leaves, so the wait always expired, the transaction failed, and the rollback deleted the prefix Live had just installed into — or restored a displaced one and discarded the new runtime.

**Every session leaked.** Closing Live, or Max, left the updater and a live `wineserver` running for the rest of the login session. It is also why `rollback` could refuse with "Live, Max, or another Wine client is running" when nothing was open.

## Changes

The payload step and both launchers end the windowless agents a session leaves behind, by exact name, from one list in `lib/lifecycle.sh`: `AbletonPushCpl.exe`, `tusbaudiocplapp.exe`, `MicrosoftEdgeUpdate.exe`. Removing the holder is enough — Wine's own processes exit once the last client goes, so the `wineserver` follows without being killed. An image that is not running is a no-op, so each Live major matches only what it has.

Ending the prefix instead would be simpler and wrong. `max9` ships its own desktop entry, so a Max and a Live routinely share one prefix; `wineserver -k` from either would take the other down mid-session.

The teardown then confirms the prefix came down rather than assuming it, and names what holds it when it did not:

```
-- the prefix is still in use, so its wineserver stays up for:
   canary.exe (pid 12345)
```

That is reported and left alone: at teardown a program we do not recognise is indistinguishable from a leftover, and users run their own programs in these prefixes. An agent a future installer invents appears there by name instead of leaking silently.

## Behaviour

| condition | before | after |
| --- | --- | --- |
| install, prefix quiet | completes | unchanged |
| install, agent resident | wait expires, rollback deletes the new prefix or restores a displaced one | agents ended, wait returns in ~4s, install completes |
| install, unrecognised process holds the prefix | same fatal path | holder reported with the `wineserver -k` command, install completes |
| app exits, agent resident | updater and `wineserver` persist for the login session | agents ended, prefix fully down |
| app exits, another app still in the prefix | no teardown existed | agents ended, nothing else signalled, server stays up for that app |
| app exits, prefix still held after the stop | — | 5s grace, holder named by Windows image and pid, nothing killed |
| launcher refuses a missing runtime or prefix | — | teardown returns early: no `wine` invoked, no prefix created |

`setup-prefix.sh`'s `[1/5]` barrier still escalates to `wineserver -k`, but only against the staging prefix it just built. `rollback.sh` and `uninstall.sh` keep their bounded waits, now sharing `ableton_prefix_wait`, and `rollback.sh`'s refusal names the processes holding the runtime instead of guessing at them. `wineserver -k` is never issued against a promoted or user-facing prefix on any path.

## Testing

60 checks in `test-installer-lifecycle.sh` and 86 in `test-pipeasio-installer.sh`, covering the expired payload wait, the named kills, a running application surviving and being named, and a refused launcher creating nothing on its way out.

Verified on Fedora 44, driven and closed through the interface. Live 12.4.2 goes `3/1/1` to `0/0/0` and Max 9.1.5 `1+3/1/1` to `0/0/0`, prefix empty both times, no `wineserver -k` on either path. A control install on `58b9656` expires the wait and rolls back where this branch completes in four seconds. Killing the updater under a running Live with seven WebView2 processes left all of them untouched, and none respawned across three further minutes.

Live 11.3.25 was installed and driven in its own prefix, and leaks a different agent by a different route: no WebView2 updater, but the Push driver's tray applet, autostarted from a Startup-folder shortcut that `explorer.exe` runs at session start. It appears as `ABLE~HEG.EXE -hide` through its 8.3 path, and `taskkill /f /im AbletonPushCpl.exe` still matches it — Wine resolves the process to its long image name. Closing Live 11 leaves the prefix empty. Both majors leak, by different agents, and the one list covers both.

## Credit — #167 by @stickyfran

@stickyfran reported this first, in #167: the `wineserver` survives the window closing, it happens from `ableton-live` **and** from standalone `max9`, and the cleanup belongs in the launchers rather than in a `.desktop` workaround. That report is what sent me looking, and the `max9` half in particular held up exactly as described.

The one thing that report could not have is *why* the server stays: a single wedged `MicrosoftEdgeUpdate.exe` holds it. That's only reason this takes a different approach. #167 calls `wineserver -k`, which ends the whole prefix. Since `max9` ships its own desktop entry, a Max and a Live routinely share one, so closing either would take the other down mid-session. Ending the named agent removes the cause instead, and the server exits on its own.

So this supersedes #167 mechanically while being downstream of it in every other sense. If it lands, #167 should close pointing here, with the find credited to @stickyfran.

### DIFF
--- a/scripts/ableton-live
+++ b/scripts/ableton-live
@@ -40,6 +40,19 @@ for lifecycle_lib in "$launcher_here/lib/lifecycle.sh" "$ABLETON_DATA_HOME/lib/l
 done
 declare -F ableton_live_running >/dev/null 2>&1 || { echo "ableton-live: lifecycle helper is missing" >&2; exit 1; }
 
+# Live exits cleanly - its extensions, its WebView2 processes and explorer all
+# go with it - but WebView2's updater does not, and one wedged agent keeps the
+# wineserver alive for the whole login session. Ending the agents by name lets
+# the server exit on its own, while a Max or a second Live in the same prefix
+# keeps it as it should.
+session_teardown()
+{
+    local rc=$?
+    ableton_session_teardown || true
+    return "$rc"
+}
+trap session_teardown EXIT
+
 WINE_ROOT="$ABLETON_WINE_ROOT"
 export WINEPREFIX="$ABLETON_WINEPREFIX"
 [ -x "$WINE_ROOT/bin/wine" ] || { echo "ableton-live: Wine runtime is missing at $WINE_ROOT" >&2; exit 1; }
--- a/scripts/lib/lifecycle.sh
+++ b/scripts/lib/lifecycle.sh
@@ -136,6 +136,76 @@ ableton_wait_for_pid_exit()
     return 1
 }
 
+# Windowless agents an install or a session starts and never stops, ended by
+# exact name so no application is touched.  MicrosoftEdgeUpdate.exe is Live 12's
+# WebView2 updater: under Wine its COM registration fails to validate, so it
+# cannot start an update worker and parks in Core::DoRun indefinitely, holding
+# the prefix open after everything else has gone.  A Max session leaves the same
+# process behind.  The Push images are the USB driver's tray applets.  An image
+# that is not running is a no-op, so callers need not know which apply.
+ableton_stop_leftover_agents()
+{
+    local runtime="${1:-$ABLETON_WINE_ROOT}" prefix="${2:-$ABLETON_WINEPREFIX}"
+    # wine builds a prefix at any path it is handed, so a caller that gave up
+    # because the runtime or prefix was missing must not create one on its way
+    # out.  Nothing is running in a prefix that does not exist either.
+    [ -x "$runtime/bin/wine" ] || return 0
+    [ -f "$prefix/system.reg" ] || return 0
+    # One invocation, not one per image: taskkill takes a list, and each wine
+    # start costs a second of a user's exit - or fifteen against a prefix that
+    # has stopped answering, three times over.
+    ableton_run_bounded 15 env WINEPREFIX="$prefix" "$runtime/bin/wine" taskkill /f \
+        /im AbletonPushCpl.exe /im tusbaudiocplapp.exe /im MicrosoftEdgeUpdate.exe \
+        >/dev/null 2>&1 || true
+    return 0
+}
+
+# Windows image name from the command line, which carries the full path where
+# /proc/PID/comm is truncated at 15 characters - short of most of them.  argv[0]
+# is read on its own NUL boundary: every Windows path has a space in it, so
+# splitting the joined command line on whitespace yields "Program".
+ableton_pid_image()
+{
+    local image
+    image="$(tr '\0' '\n' < "/proc/$1/cmdline" 2>/dev/null | head -n 1)"
+    image="${image##*\\}"
+    image="${image##*/}"
+    # A process that exits while it is being reported leaves nothing to read.
+    [ -n "$image" ] || image="$(cat "/proc/$1/comm" 2>/dev/null || true)"
+    printf '%s\n' "${image:-unknown}"
+}
+
+# End a session: stop the agents it leaves behind, then confirm the prefix
+# actually came down rather than assume it.  Wine's own processes exit once the
+# last client goes, so a prefix still busy after that grace period is being held
+# by something real - a Max, a second Live, or a program the user started in this
+# prefix themselves.  That is reported and left alone: at this point we cannot
+# tell a user's program from a leftover, and ending the session would take it
+# with us.  Returns non-zero when the prefix is still held.
+ableton_session_teardown()
+{
+    local runtime="${1:-$ABLETON_WINE_ROOT}" prefix="${2:-$ABLETON_WINEPREFIX}"
+    local seconds="${3:-5}" i pid
+    seconds="$(ableton_timeout_value "$seconds" teardown-grace 1 60)" || return 2
+    # Nothing to end, and nothing to wait for.  Checked first because taskkill is
+    # itself a wine process: on an empty prefix it starts a server and a pair of
+    # services that outlive the grace period below, and the teardown would then
+    # report the session it started as the thing holding the prefix.
+    ableton_prefix_busy || return 0
+    ableton_stop_leftover_agents "$runtime" "$prefix" || true
+    for ((i=0; i<seconds*10; i++)); do
+        ableton_prefix_busy || return 0
+        sleep 0.1
+    done
+    ableton_prefix_busy || return 0
+    printf -- '-- the prefix is still in use, so its wineserver stays up for:\n' >&2
+    while IFS= read -r pid; do
+        [ -n "$pid" ] || continue
+        printf '   %s (pid %s)\n' "$(ableton_pid_image "$pid")" "$pid" >&2
+    done < <(ableton_prefix_pids)
+    return 1
+}
+
 # Wait for every process in the prefix to exit.  Bounded: a resident that outlives
 # the command that started it holds the server open indefinitely.
 ableton_prefix_wait()
--- a/scripts/lib/lifecycle.sh
+++ b/scripts/lib/lifecycle.sh
@@ -2,11 +2,12 @@
 # Prefix- and runtime-scoped Wine lifecycle inspection.  No process is selected
 # by a global image-name match: both /proc/PID/exe and WINEPREFIX must match.
 
+# Definitions only: callers resolve the configuration themselves, because the
+# command line reaches some of them after this file is sourced.
 _ableton_lib_dir="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
 if ! declare -F ableton_config_init >/dev/null 2>&1; then
     . "$_ableton_lib_dir/config.sh"
 fi
-ableton_config_init
 
 ableton_pid_has_env()
 {
@@ -133,6 +134,38 @@ ableton_wait_for_pid_exit()
         sleep 0.1
     done
     return 1
+}
+
+# Wait for every process in the prefix to exit.  Bounded: a resident that outlives
+# the command that started it holds the server open indefinitely.
+ableton_prefix_wait()
+{
+    local runtime="${1:-$ABLETON_WINE_ROOT}" prefix="${2:-$ABLETON_WINEPREFIX}"
+    ableton_run_bounded 60 env WINEPREFIX="$prefix" \
+        "$runtime/bin/wineserver" -w >/dev/null 2>&1
+}
+
+# Wait, and on timeout end every process in the prefix and wait again.  The stop is
+# indiscriminate, so this is only for a prefix the caller owns outright - a staging
+# prefix it just built, or one it is about to remove.  On a prefix the user can
+# reach, name the images to end instead and take ableton_prefix_wait's verdict as
+# advisory.  wineserver -k shuts down through SIGINT, so the registry is saved.
+# Returns 1 when a straggler survived the stop, and the wait's own exit code when
+# the wait could not run at all.
+ableton_prefix_quiesce()
+{
+    local runtime="${1:-$ABLETON_WINE_ROOT}" prefix="${2:-$ABLETON_WINEPREFIX}" rc=0
+    ableton_prefix_wait "$runtime" "$prefix" || rc=$?
+    if [ "$rc" -eq 0 ]; then
+        return 0
+    fi
+    if [ "$rc" -ne 124 ] && [ "$rc" -ne 137 ]; then
+        return "$rc"
+    fi
+    echo "-- a leftover background program is holding the prefix open; stopping it"
+    ableton_run_bounded 20 env WINEPREFIX="$prefix" \
+        "$runtime/bin/wineserver" -k >/dev/null 2>&1 || true
+    ableton_prefix_wait "$runtime" "$prefix" || return 1
 }
 
 ableton_stop_prefix()
--- a/scripts/max9
+++ b/scripts/max9
@@ -13,6 +13,17 @@ for lib in "$here/lib/lifecycle.sh" "$ABLETON_DATA_HOME/lib/lifecycle.sh"; do
 done
 declare -F ableton_max_pids >/dev/null 2>&1 || { echo "max9: lifecycle helper is missing" >&2; exit 1; }
 
+# Max exits cleanly - its node.exe helpers and console host go with it - but a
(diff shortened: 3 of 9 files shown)


scripts/ableton-live
```diff
@@ -40,6 +40,19 @@ for lifecycle_lib in "$launcher_here/lib/lifecycle.sh" "$ABLETON_DATA_HOME/lib/l
...
+# wineserver alive for the whole login session. Ending the agents by name lets
+# the server exit on its own, while a Max or a second Live in the same prefix
+# keeps it as it should.
+session_teardown()
+{
+    local rc=$?
+    ableton_session_teardown || true
+    return "$rc"
+}
+trap session_teardown EXIT
```
shibco: Gate the teardown on the launcher owning the bring-up. Completed by `session_started=1` beside `launch_lock_held=1` (after line 285) - a suggestion cannot reach that line.

```suggestion
# Live exits cleanly - its extensions, its WebView2 processes and explorer all
# go with it - but WebView2's updater does not, and one wedged agent keeps the
# wineserver alive for the whole login session. Ending the agents by name lets
# the server exit on its own, while a Max or a second Live in the same prefix
# keeps it as it should.
# Armed for every exit, acting only once the launcher owns the bring-up: a
# refused or warm launch must not reach into a prefix another session is using.
session_started=""
session_teardown()
{
    local rc=$?
    [ -z "$session_started" ] || ableton_session_teardown || true
    return "$rc"
}
trap session_teardown EXIT
```

shibco (commented):
Doublepost

scripts/max9
```diff
@@ -13,6 +13,17 @@ for lib in "$here/lib/lifecycle.sh" "$ABLETON_DATA_HOME/lib/lifecycle.sh"; do
...
+# Max exits cleanly - its node.exe helpers and console host go with it - but a
+# Max session leaves WebView2's updater behind exactly as a Live session does,
+# and that one agent holds the wineserver for the rest of the login session.
+session_teardown()
+{
+    local rc=$?
+    ableton_session_teardown || true
+    return "$rc"
+}
+trap session_teardown EXIT
```
shibco: Same gate as ableton-live. Completed by `session_started=1` after `printf 'max %s\n' "$$" >&9` (line 51) - a suggestion cannot reach that line.

```suggestion
# Max exits cleanly - its node.exe helpers and console host go with it - but a
# Max session leaves WebView2's updater behind exactly as a Live session does,
# and that one agent holds the wineserver for the rest of the login session.
# Armed for every exit, acting only once the launcher owns the bring-up: a
# refused launch must not reach into a prefix another session is using.
session_started=""
session_teardown()
{
    local rc=$?
    [ -z "$session_started" ] || ableton_session_teardown || true
    return "$rc"
}
trap session_teardown EXIT
```

scripts/lib/lifecycle.sh
```diff
@@ -135,6 +136,108 @@ ableton_wait_for_pid_exit()
...
+# actually came down rather than assume it.  Wine's own processes exit once the
+# last client goes, so a prefix still busy after that grace period is being held
+# by something real - a Max, a second Live, or a program the user started in this
+# prefix themselves.  That is reported and left alone: at this point we cannot
+# tell a user's program from a leftover, and ending the session would take it
+# with us.  Returns non-zero when the prefix is still held.
+ableton_session_teardown()
+{
+    local runtime="${1:-$ABLETON_WINE_ROOT}" prefix="${2:-$ABLETON_WINEPREFIX}"
+    local seconds="${3:-5}" i pid
```
shibco:
```suggestion
    local seconds="${3:-5}" deadline pid
```
`i` goes with the iteration count below.

scripts/lib/lifecycle.sh
```diff
@@ -135,6 +136,108 @@ ableton_wait_for_pid_exit()
...
+    # Nothing to end, and nothing to wait for.  Checked first because taskkill is
+    # itself a wine process: on an empty prefix it starts a server and a pair of
+    # services that outlive the grace period below, and the teardown would then
+    # report the session it started as the thing holding the prefix.
+    ableton_prefix_busy || return 0
+    ableton_stop_leftover_agents "$runtime" "$prefix" || true
+    for ((i=0; i<seconds*10; i++)); do
+        ableton_prefix_busy || return 0
+        sleep 0.1
+    done
```
shibco: Each `ableton_prefix_busy` walks all of /proc, which costs whole seconds on a loaded box; fifty iterations cost minutes, not the five seconds the grace names (108.8s measured with wine stubs that exit instantly). A deadline keeps the grace at its stated size.

```suggestion
    deadline=$((SECONDS + seconds))
    while [ "$SECONDS" -lt "$deadline" ]; do
        ableton_prefix_busy || return 0
        sleep 0.1
    done
```

scripts/ableton-live
shibco, line 54: The trap lands above the guards, so a refusal runs the teardown too. Gated on a flag the launcher sets once it owns the session. The completing line sits outside the diff: `session_started=1` right after `printf 'live %s\n' "$$" >&9` (line 285). `launch_lock_held` looked like a ready-made gate, but the supervisor resets it to 0 once the prefix exposes Live, so it is gone again by session end.

```suggestion
# Live exits cleanly - its extensions, its WebView2 processes and explorer all
# go with it - but WebView2's updater does not, and one wedged agent keeps the
# wineserver alive for the whole login session. Ending the agents by name lets
# the server exit on its own, while a Max or a second Live in the same prefix
# keeps it as it should.
# Armed before the guards, acting only once this launcher owns a session: a
# refusal - missing runtime, prefix, or Live, or another launcher holding the
# bring-up lock - must not reach into a prefix another session is using.
session_started=""
session_teardown()
{
    local rc=$?
    [ -z "$session_started" ] || ableton_session_teardown || true
    return "$rc"
}
trap session_teardown EXIT
```

scripts/max9
shibco, line 25: Same gate as ableton-live. The completing line, outside the diff: `session_started=1` right after `printf 'max %s\n' "$$" >&9` (line 51).

```suggestion
# Max exits cleanly - its node.exe helpers and console host go with it - but a
# Max session leaves WebView2's updater behind exactly as a Live session does,
# and that one agent holds the wineserver for the rest of the login session.
# Armed before the guards, acting only once this launcher owns a session: a
# refusal - missing runtime, prefix, or Max, or another launcher holding the
# bring-up lock - must not reach into a prefix another session is using.
session_started=""
session_teardown()
{
    local rc=$?
    [ -z "$session_started" ] || ableton_session_teardown || true
    return "$rc"
}
trap session_teardown EXIT
```

scripts/lib/lifecycle.sh
shibco, line 188: `i` gives way to `deadline` for the wall-clock grace below.

```suggestion
    local seconds="${3:-5}" deadline pid
```

shibco (commented):
Reviewed on a trial merge onto current main (0460dd4). The merge is clean and both suites pass on it, 65/65 and 86/86. The installer change works: an expired payload wait completes the install, names the straggler, and leaves the promoted prefix alone. This fixes #213 and #215.

One defect blocks the merge. Both launchers arm the EXIT trap before their guards, so a refused launch runs the teardown into a busy prefix. Repro on the merged tree: a Live from another session holds the prefix, and max9 runs with Max missing:

```
max9: no Max 9 installation in .../prefix
wine taskkill /f /im AbletonPushCpl.exe /im tusbaudiocplapp.exe /im MicrosoftEdgeUpdate.exe
-- the prefix is still in use, so its wineserver stays up for:
   Ableton Live 12 Suite.exe (pid 247569)
exit=1  elapsed=108.8s
```

This kills agents inside the other session's prefix, reports that session as the blocker, and unfortunately takes 109 seconds. A failed double launch does the same during bring-up. The suite passes today because every refusal fixture uses an idle prefix. The delay is the grace loop: 50 iterations, each a full walk of /proc.

scripts/lib/lifecycle.sh
shibco, line 199: Each `ableton_prefix_busy` pass walks all of /proc; at about two seconds per pass on a loaded system, fifty passes stretch this five-second grace to minutes. The 108.8s in the refusal repro is this loop, with wine stubs that exit instantly. The same cost lands on a legitimate exit that leaves the prefix held, e.g. closing Live while a Max is still in it.

```suggestion
    # A wall-clock deadline, not an iteration count: each ableton_prefix_busy
    # pass walks all of /proc and can cost seconds, and fifty passes turned
    # this grace into minutes on a busy system.
    deadline=$((SECONDS + seconds))
    while [ "$SECONDS" -lt "$deadline" ]; do
        ableton_prefix_busy || return 0
        sleep 0.1
    done
```

scripts/test-installer-lifecycle.sh
```diff
@@ -527,21 +527,204 @@ grep -q 'appears to be Live 11' "$base/err" || fail "payload mismatch is explici
...
+EOF
+printf '#!/bin/sh\nexit 0\n' > "$base/runtime/bin/wineserver"
+chmod +x "$base/runtime/bin/wine" "$base/runtime/bin/wineserver"
+if run_isolated "$base" env ABLETON_WINE_ROOT="$base/runtime" \
+    ABLETON_WINEPREFIX="$base/absent-prefix" bash "$here/ableton-live" \
+    >"$base/out" 2>"$base/err"; then
+    fail "launcher accepts a missing prefix"
+fi
+[ ! -e "$base/absent-prefix" ] || fail "teardown builds a prefix the launcher refused to use"
+ok "a launcher that refuses a missing prefix creates nothing on its way out"
```
shibco, line 726: Pins the refusal path against a busy prefix, mirroring the max-coexist fixture. Fails on the current branch (taskkill fired, prefix reported); passes with the gate and the two flag lines in place.

```suggestion
ok "a launcher that refuses a missing prefix creates nothing on its way out"

# A refusal must not reach into the prefix: with another session's client still
# running, a launcher that declines to start ends nothing and reports nothing.
base="$(new_env refusal-busy-prefix)"
mkdir -p "$base/runtime/bin" "$base/prefix"
cat > "$base/runtime/bin/wine" <<'EOF'
#!/bin/sh
printf 'wine %s\n' "$*" >> "${ABLETON_TEST_LOG:?}"
EOF
printf '#!/bin/sh\nexit 0\n' > "$base/runtime/bin/wineserver"
chmod +x "$base/runtime/bin/wine" "$base/runtime/bin/wineserver"
printf 'registry\n' > "$base/prefix/system.reg"
cp /bin/sleep "$base/runtime/bin/wine-client"
env WINEPREFIX="$base/prefix" bash -c 'exec -a "C:\\ProgramData\\Ableton\\Live 12 Suite\\Program\\Ableton Live 12 Suite.exe" "$1" 600' _ "$base/runtime/bin/wine-client" &
holder_pid=$!
sleep 0.1
: > "$base/wine.log"
if run_isolated "$base" env ABLETON_WINE_ROOT="$base/runtime" ABLETON_WINEPREFIX="$base/prefix" \
    ABLETON_TEST_LOG="$base/wine.log" bash "$here/max9" >"$base/out" 2>"$base/err"; then
    fail "max9 without a Max installation fails"
fi
kill "$holder_pid" 2>/dev/null
wait "$holder_pid" 2>/dev/null || true
! grep -q 'taskkill' "$base/wine.log" || fail "a refused launch ends agents in a busy prefix"
! grep -q 'prefix is still in use' "$base/err" || fail "a refused launch reports another session's prefix"
ok "a refused launch leaves a busy prefix alone"
```

PR #222, opened 2026-08-16
Feat/settle progress and lint

ClickSentinel:
Stacked on #220 — review that first; this assumes its teardown exists.

## Behaviour

**Closing Live, nothing else in the prefix.** The terminal comes straight back; the helper finishes and the wineserver exits with it.

```
-- Ableton Live closed; a background helper is still finishing and will quit on its own
```

**Closing Live with something else still in the prefix.** Nothing is killed.

```
-- Ableton Live closed. Other unknown processes were left running:
   reaper.exe (pid 412402)
-- Ableton-Linux helpers close themselves once the prefix is free. To kill the prefix forcefully instead:
   WINEPREFIX=/home/you/.wine-ableton /home/you/.local/opt/wine-d2d1-nspa-11.13/bin/wineserver -k
```

**Waiting during an install.** Was a silent minute, which reads as a hang.

```
   still waiting for the prefix to settle (15s): MicrosoftEdgeUpdate.exe
```

**A busy prefix at the end of an install.** The install is done, exit.

```
-- the install is complete. A program in the prefix is still running and
   was left alone:
   reaper.exe (pid 412402)
-- nothing needs to be done. To end it anyway, close it or run:
   WINEPREFIX=/home/you/.wine-ableton /home/you/.local/opt/wine-d2d1-nspa-11.13/bin/wineserver -k
```

**A launcher that refuses to start** — missing runtime, no Max installed, bad `ABLETON_LIVE_VERSION` — now says only that, and touches nothing. Previously it ran teardown against whatever session was already using the prefix.

Three defects from a review of `lib/lifecycle.sh`

Each has a test that fails without its fix, confirmed by reverting the fix and watching it go red.

- **EXIT trap armed above the preflight guards.** A refused launch `taskkill`ed three agents inside a running Live, then advised `wineserver -k` on the prefix the user was working in.
- **`ableton_pid_image` opened `/proc/PID/cmdline` before stderr was silenced.** Redirections apply left to right, so a pid exiting mid-scan printed a shell error — from the EXIT trap, as Live closes.
- **The `/proc` walkers hardcoded the configured prefix** while callers passed one explicitly, so setup-prefix.sh's staging window reported on the user's prefix. They now take runtime/prefix defaulting to the globals. The operations always used the prefix they were given; this was a reporting defect.

Also: the ticking wait is now the bounded wait, backgrounded and watched, not a second copy with its own 60s bound to drift. And `.shellcheckrc` lets ShellCheck follow the sourced libraries — six SC1091s become none, surfacing one previously invisible diagnostic.

## Testing

Suites at 64 and 86, green on the host and on a Fedora 44 guest — the distro both issues were filed from — against a checksum-verified identical tree.

By hand there, on a real Live 12 session:

- Closed Live with six `msedgewebview2.exe` running; all six went with it, the wineserver stayed up while `learnheal.exe` finished, gone by +90s. #215's symptom.
- A session ending with an updater parked killed exactly one process, `MicrosoftEdgeUpdate.exe`. Live, WebView2 and `Ableton Index.exe` were named and left alive.
- A refused launch with the prefix busy left the holder list byte-identical and printed nothing.
- Not reproduced: #213's wedged updater. Omaha parks one on its own schedule, not on every launch; that comparison was made earlier against `58b9656`.

## Not addressed here

- `uninstall.sh:458` — `ableton_prefix_busy && ableton_stop_prefix` under `set -euo pipefail`, and `ableton_stop_prefix` ends on `! ableton_prefix_busy`. One surviving process exits the script silently, after the trap is cleared, skipping PipeASIO unregister, shortcut restore, MIME cleanup and runtime removal. Byte-identical on `main`, so it wants its own change.
- The SIGTERM loop lacks the identity guard its SIGKILL counterpart has. Same story.

## Credits

- **@realitymolder** — #212, source of the `.shellcheckrc` and progress-tick ideas. Ports of the ideas, not the code: that PR's `settle()` is entangled with a `setup-prefix.sh` predating the installer transactions, and its companion refusal matches `pgrep -af` across every prefix on the machine.
- **@Version33** — #27, the Flake work #212 was split out of.
- **@yioannides** (#213) and **@K4M11111** (#215) — the reports that started this.

### DIFF
--- a/.shellcheckrc
+++ b/.shellcheckrc
@@ -1,0 +1,2 @@
+external-sources=true
+source-path=SCRIPTDIR
--- a/scripts/ableton-live
+++ b/scripts/ableton-live
@@ -40,11 +40,8 @@ for lifecycle_lib in "$launcher_here/lib/lifecycle.sh" "$ABLETON_DATA_HOME/lib/l
 done
 declare -F ableton_live_running >/dev/null 2>&1 || { echo "ableton-live: lifecycle helper is missing" >&2; exit 1; }
 
-# Live exits cleanly - its extensions, its WebView2 processes and explorer all
-# go with it - but WebView2's updater does not, and one wedged agent keeps the
-# wineserver alive for the whole login session. Ending the agents by name lets
-# the server exit on its own, while a Max or a second Live in the same prefix
-# keeps it as it should.
+# Live exits cleanly; WebView2's updater does not, and one wedged agent keeps the
+# wineserver alive for the whole login session.
 session_teardown()
 {
     local rc=$?
--- a/scripts/ableton-live
+++ b/scripts/ableton-live
@@ -45,7 +45,7 @@ declare -F ableton_live_running >/dev/null 2>&1 || { echo "ableton-live: lifecyc
 session_teardown()
 {
     local rc=$?
-    ableton_session_teardown || true
+    ABLETON_SESSION_LABEL="Ableton Live" ableton_session_teardown || true
     return "$rc"
 }
 trap session_teardown EXIT
--- a/scripts/ableton-live
+++ b/scripts/ableton-live
@@ -40,6 +40,19 @@ for lifecycle_lib in "$launcher_here/lib/lifecycle.sh" "$ABLETON_DATA_HOME/lib/l
 done
 declare -F ableton_live_running >/dev/null 2>&1 || { echo "ableton-live: lifecycle helper is missing" >&2; exit 1; }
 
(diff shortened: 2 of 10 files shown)


shibco:
In isolation, #222 is mergeable and passes the full suites against current main.

But #230 does supersedes both #220 and #222’s core policies of having a 15-second wait + direct confirmation + whole-prefix shutdown.

Therefore, we should hold off until we finish untangling this mess. lmao.

PR #227, opened 2026-08-18
no longer extracting in memory, what the fuck was i thinking?

### LINKED ISSUE #226: Payload extract write error when disk is not full
<kbd>[lee](https://discord.com/channels/1307779762189504602/1539258800047919114/1539258800047919114)</kbd> [lee](https://discord.com/channels/1307779762189504602/1539258800047919114/1539258800047919114)  `BOT`

```
OK: prefix ready at /home/wyzux/.wine-ableton
-- extracting Live installer payload (bounded to 900s; extracted filenames show progress)
Archive:  /home/wyzux/Downloads/ableton_live_suite_12.4.3_64.zip
  inflating: /tmp/ableton-live-installer.q5v3CS/Ableton Live 12 Suite Installer.exe  
  inflating: /tmp/ableton-live-installer.q5v3CS/Ableton Live 12 Suite Installer-1.bin  
  inflating: /tmp/ableton-live-installer.q5v3CS/Ableton Live 12 Suite Installer-2.bin  
  inflating: /tmp/ableton-live-installer.q5v3CS/Ableton Live 12 Suite Installer-3.bin  
/tmp/ableton-live-installer.q5v3CS/Ableton Live 12 Suite Installer-3.bin:  write error (disk full?).  Continue? (y/n/^C) 
```
```
 wyzux@moon  ~/Downloads  df -h /tmp
Filesystem      Size  Used Avail Use% Mounted on
tmpfs           7.6G  6.1G  1.6G  80% /tmp
```

shibco:
the first version of the installer just shot out the extracted Ableton installer into the same dir.
in the second version i thought i'd be clever in case someone was running the installer + ableton install file from read only media, the idea was to make it totally resilient depending on the environment.
and of course that led to MULTIPLE people being unable to extract Live
this is so stupid.

anyway its fixed
also the installer now won't re-extract over the top of an already-extracted copy of the live installer.

Fixes #226

### DIFF
--- a/scripts/installer.sh
+++ b/scripts/installer.sh
@@ -748,14 +748,21 @@ install_live_payload()
     local installer="$live_payload" unpack="" lower flags=() timeout_secs extract_timeout status=0 tray seed_reg=""
     lower="$(basename "$installer" | tr '[:upper:]' '[:lower:]')"
     if [[ "$lower" = *.zip ]]; then
-        unpack="$(mktemp -d "${TMPDIR:-/tmp}/ableton-live-installer.XXXXXX")"
+        unpack="$(dirname "$installer")/ableton-live-installer.d"
+        if [ -e "$unpack/.extracted" ]; then
+            echo "-- reusing the Live installer payload already extracted at $unpack"
+        else
+            live_unpack="$unpack"; cleanup_live_unpack || return 1
+            mkdir -p -- "$unpack"
+            extract_timeout="$(ableton_timeout_value "${ABLETON_PAYLOAD_EXTRACT_TIMEOUT:-900}" ABLETON_PAYLOAD_EXTRACT_TIMEOUT 60 7200)"
+            echo "-- extracting Live installer payload (bounded to ${extract_timeout}s; extracted filenames show progress)"
+            if command -v unzip >/dev/null 2>&1; then ableton_run_bounded "$extract_timeout" unzip "$installer" -d "$unpack"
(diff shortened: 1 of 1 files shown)


ClickSentinel:
`.extracted` doesn't seem to record which zip wrote it, and the Live version check reads the zip file name. Passing `ableton_live_lite_11.3.35_64.zip` resolved `Ableton Live 12 Suite Installer.exe`. you could write the zip's name and size into `.extracted`, and compare possibly. 

we could do other folder check things like perms... and let people send it elsewhere with an ABLETON_PAYLOAD_UNPACK_DIR setting or something. on a read-only folder the run breaks at a `mkdir: Permission denied` and then you get rollback message.

we could check free space first too? and run the extractor with `</dev/null`. live 12.4.3 trial becomes 6.48 GB next to its 4.45 GB zip, and unzip then waits on its "Continue" for 900 seconds lolol. don't ask how I found know this. 

Otherwise lol funny name.

shibco:
Thanks @ClickSentinel 

1. .extracted now records which zip produced it (name + size), so a leftover dir from one zip no longer gets reused
2. `ABLETON_PAYLOAD_UNPACK_DIR` can redirect the unpack location, and my hyper-clever attempt at mitigating this automatically is now a DIR and a read-only target that now fails early with a clear message instead of `mkdir: Permission denied`
3. Added a free-space pre-check (~1.5× the zip) and the extractors read from /dev/null so unzip can't choke to death on a `"replace? (y/n)"` prompt.

ClickSentinel (approved):

PR #230, opened 2026-08-19
Refined installer

### LINKED ISSUE #213: Installer fails on fresh re-install (after uninstalling)

shibco:
The previous iteration of this installer waited 60 seconds for background processes we know will never quit by themselves, and then failed entirely. This would happen reliably despite a completed runtime, prefix and Ableton Live install. The rollback then deleted that finished install. On an upgrade it restored the old prefix instead and discarded the new runtime.

These changes fix #213 and #215, both of which are the result of this overly cautious behaviour.

- Cut install time delays to 15 seconds or less, so a slow background process no longer fails the install.
- Prompt the user in the CLI during an Installer flow to ask them directly whether to terminate running Wine instances. This can be automatically agreed to via a `--yes` flag.
- Shut down the prefix instead of listing processes, including Live 11 specific weirdness.
- Dropped the two hardcoded process names, which is now redundant thanks to the above changes.
- Made six variables in the payload step local, as they were global by mistake.
- Retired the old PR #182 Link binary by standardising the Link install with the helpers used across the installer.
- Removed stale installer code from the refactor from last week.
- Added README and TROUBLESHOOTING entries for the new user prompting question (to kill running Wine processes) and how to mitigate the Hyprland stall.
- Removed two additional race checks.

### DIFF
--- a/README.md
+++ b/README.md
@@ -67,6 +67,10 @@ For best results, double-check the name of the Ableton installer you have downlo
 
 Once started, this is a mostly automated process.
 
+Near the end, the installer may ask whether to close a background program that
+Ableton's installer left running. Press Enter to close it. Live is already
+installed by then, so either answer keeps it.
+
 ### Running Live
 
 Start Ableton Live the way you normally start applications on your Linux OS.
--- a/TROUBLESHOOTING.md
+++ b/TROUBLESHOOTING.md
@@ -16,6 +16,7 @@ have already fixed your issue.
 
 - [Installation and updates](#installation-and-updates)
   - [The installer does not finish after Live installs](#the-installer-does-not-finish-after-live-installs)
+  - [The Ableton installer window never progresses or shows graphical corruption on Hyprland](#the-ableton-installer-window-never-progresses-or-shows-graphical-corruption-on-hyprland)
   - [The installer says PipeWire is too old](#the-installer-says-pipewire-is-too-old)
 - [Live versions and launching](#live-versions-and-launching)
   - [The launcher finds more than one Live installation](#the-launcher-finds-more-than-one-live-installation)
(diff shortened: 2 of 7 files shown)


ClickSentinel:
Ok so this is really good but there are a couple of thing I did differently in 220 that are worth your call. 

In this version we would automatically kill the process when the wait expires, but my approach was to identify the dumb services that keep running. And it's only 1 additional right now: M*crosoft Edge Update. When we kill it, the server automatically shuts down gracefully after a little bit. I agree about the list becoming redundant, but we could probably do this both ways.

```bash
ableton_stop_leftover_agents          # AbletonPushCpl / tusbaudiocplapp / MicrosoftEdgeUpdate
if ! ableton_prefix_wait; then        # 60s
    echo "-- the install is complete. A program in the prefix is still running and"
    echo "   was left alone:"
    for holder in $(ableton_prefix_pids); do
        printf '   %s (pid %s)\n' "$(ableton_pid_image "$holder")" "$holder"
    done
    echo "-- nothing needs to be done. To end it anyway, close it or run:"
    printf '   WINEPREFIX=%s %s/bin/wineserver -k\n' \
        "$ABLETON_WINEPREFIX" "$ABLETON_WINE_ROOT"
fi
```
We should probably discuss a little more.

PR #231, opened 2026-08-19
Consolidate/installer lifecycle

ClickSentinel:
## Consolidate the installer and session lifecycle fixes from #220, #222 and #230

All three fix the same failure — `install_live_payload`'s wineserver wait expiring, the transaction failing, and rollback deleting a Live that had already installed (#213, #215) — from opposite ends. They split cleanly: #230 owns the installer's payload step and the dead code around it, #220/#222 own the session teardown and the shared `/proc` walkers. A trial merge conflicts in one hunk, `install_live_payload` itself.

The payload step now does both rather than one. The agents an install leaves behind are ended by name first, because that removes the cause and needs nobody's permission; only what survives that is reported. Ending the prefix is offered, defaulting to leaving it, and `--yes` does not answer it — `--yes` covers the installer's own files, and a Max sharing the prefix is not one of them.

Also fixed here: both launchers gate the EXIT teardown on owning a session, so a refusal no longer reaches into another session's prefix; `ableton_prefix_quiesce` distinguishes a straggler that survived the stop from a wait that could not run, which `setup-prefix.sh` needs; `uninstall.sh` no longer exits silently when a process outlives the stop, skipping the PipeASIO unregister, shortcut restore, MIME cleanup and runtime removal; `ableton_stop_prefix` re-checks identity before SIGTERM as it already did before SIGKILL. Two further bugs came from running a real session rather than stubs: `AbletonAudioCpl.exe` was missing from the agent list, and an agent the teardown had just ended was reported as the user's own program because `taskkill` returns before its targets exit.

**Testing.** 72 installer lifecycle checks and 84 focused PipeASIO checks. On Fedora 44: both suites green in the guest, the `/proc` walkers verified against 27 real Wine processes, and the full `installer.sh` transaction completed in 265s with Live installed, the prefix's registry intact and nothing left holding it. Two limits worth stating: the real-Wine evidence is Fedora only, and the expiry branch — the actual fix — did not fire on that run, so it is covered by the suite and by two direct probes rather than by a live wedged updater.

Supersedes #220, #222 and #230, which should close pointing here. Credit: @stickyfran (#167) for the session-leak report and the launcher-side diagnosis, @shibco for the installer work this builds on, @realitymolder (#212) and @Version33 (#27) for the `.shellcheckrc` and progress-tick ideas, @yioannides (#213) and @K4M11111 (#215) for the reports.

### DIFF
--- a/.shellcheckrc
+++ b/.shellcheckrc
@@ -1,0 +1,2 @@
+external-sources=true
+source-path=SCRIPTDIR
--- a/README.md
+++ b/README.md
@@ -67,6 +67,10 @@ For best results, double-check the name of the Ableton installer you have downlo
 
 Once started, this is a mostly automated process.
 
+Near the end, the installer may ask whether to close a background program that
+Ableton's installer left running. Press Enter to close it. Live is already
+installed by then, so either answer keeps it.
+
 ### Running Live
 
 Start Ableton Live the way you normally start applications on your Linux OS.
--- a/TROUBLESHOOTING.md
+++ b/TROUBLESHOOTING.md
@@ -16,6 +16,7 @@ have already fixed your issue.
 
 - [Installation and updates](#installation-and-updates)
   - [The installer does not finish after Live installs](#the-installer-does-not-finish-after-live-installs)
+  - [The Ableton installer window never progresses or shows graphical corruption on Hyprland](#the-ableton-installer-window-never-progresses-or-shows-graphical-corruption-on-hyprland)
   - [The installer says PipeWire is too old](#the-installer-says-pipewire-is-too-old)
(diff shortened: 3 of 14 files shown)


shibco:
This is so close.

1. A failed shutdown can delete an in-use runtime and prefix. `scripts/uninstall.sh:464` only logs when `ableton_stop_prefix` fails. With `--delete-prefix`, we don't check runs and subsequently both trees are recursively removed. Mark the uninstall partial or recheck process ownership before deletion.
2. In `scripts/lib/lifecycle.sh:164`, if something from Wine remains after the grace period, it is filtered out and reported as something that “will quit on its own” but doesn't  that include our nemesis, `MicrosoftEdgeUpdate.exe`?

PR #237, opened 2026-08-20
Flake/rederive

ClickSentinel:
This is @Version33's flake from #27, re-derived onto current `main` instead of merged into it. The general fixes are left out and the merge regressions are dropped.

`nix run .#setup-prefix` makes the Wine prefix, `nix run .` launches Live. `.#wine`, `.#check-ntsync`, `.#setup-realtime`, `.#setup-link` and `.#audio-report` cover the tarball's extra steps.

You need `experimental-features = nix-command flakes` in your nix.conf, and only `x86_64-linux` is supported.

**Credit**

The packaging is @Version33's and I barely touched it. `flake.lock` and `nix/ableton-linkd.nix` are byte-identical to #27, and `flake.nix` differs only by the added `.#audio-report` app. `nix/pipeasio.nix`, `nix/wine.nix` and `nix/ableton-wine.nix` differ by the fixes below. The build gates do a little more: `--with-gstreamer` turns a drop into a configure error, and the `env -i` smoke gate came with a diagnosis of why nix's ASLR-off builders made wine segfault.

- @realitymolder did the split into #212 and the rebase work behind it.
- @Immabed found the store-path `.desktop` failure that the runtime link exists to fix, and then the missing settings panel on this PR.
- @Username404-59 found the `unzip` collision, which is why the kit-private tools stay out of `$out/bin`.
- @shibco's four review items on #27 shaped most of the work.

**Different from #27**

Dropped, because `main` already does them another way: `settle()`, the `ls 00*.patch` audit check, the PipeASIO 1.2.2 comment, and the deletion of `shortcut-hold.sh`'s `SC2034` directive.

Fixed:

- Installer selection sorts on a version key, so an edition can't outrank a version any more.
- The Live install uses `ABLETON_LIVE_INSTALL_TIMEOUT` (3600s) instead of the 300s wine-command bound.
- `[6/6]` is gated on `ABLETON_PREFIX_MANAGED`, so it does nothing during a `.run` install.
- Silent flags come from the installer engine, and the payload has to hold exactly one `.exe`.
- `runtime-link.sh` resolves `$ABLETON_DATA_HOME` instead of a hardcoded `$HOME`.
- Handler entries that are symlinks into the store get left alone for nix or home-manager to own.
- The package ships `update-desktop-database` and `xdg-mime`, so the entries actually become the handler for `ableton://` links.
- The Qt settings panel is built and staged. Live's Hardware Setup button doesn't open a window itself: patch `pipeasio/0011` looks `pipeasio-settings` up on PATH and spawns it, so with `BUILD_SETTINGS_PANEL=OFF` a nix user only got the dialog telling them to edit config.ini by hand. The container build has defaulted it ON the whole time.
- CI needs no `verify-installer-payload.py` change, because `runtime-link.sh` doesn't go into the `.run` kit at all. The tarball never uses the link.

**Fixed after review** (`cb5fbe9`, `8d20bf3`)

- The `00*.patch` glob I dropped from #27's audit check had also reached the flake's own provenance stamp, where it does more damage: `wine.nix` applies the series from `SERIES.sha256`, so at patch 0100 it would apply that patch while the stamp omitted it, and the audit would then reject a tree whose only fault is the stamp. Now `[0-9][0-9][0-9][0-9]-*.patch`, matching `container-build.sh`.
- Patches were applied with bare `patch(1)`, which defaults to fuzz 2 and places a drifted hunk at exit 0, where `container-build.sh`'s `git am --3way` never does. Now `--fuzz=0 -N --no-backup-if-mismatch`; all 97 apply unchanged.
- `audio-report.sh` was not packaged at all, though the bug-report template asks every reporter to run it. Staged, gated, and exposed as `.#audio-report`. Found by diffing `make-installer.sh`'s kit against what the package installs; that diff is now clean apart from the tarball's own transaction tools.
- `newest_live_zip` let a `find` that exits 1 reach an unguarded assignment, so under `set -e` an unreadable `~/Proprietary` ended the whole prefix setup with no message.
- Folding three rpath checks into `rpath_check` moved the `readelf` pipeline out of an `if` condition, where `set -e` is suppressed, into a bare assignment where it is not. A missing file now reports a failure instead of stopping the audit mid-run.
- `runtime-link.sh`'s fast path carried an `[ -e "$link" ]` test that cannot fire, since the path is only reached when the target already resolved. Removed with the comment that claimed it.

**Size**

18 files, +2130/-79. 1614 of that is new files, the rest is +516/-79 across ten existing ones.

**Testing**

NixOS: `nix flake check` passes, cold build 14m53s including Wine from source, audit green at 163 checks with 6 provenance records skipped under `ABLETON_AUDIT_PROFILE=nix`. The `release` profile fails on exactly those six, so a nix build can't get shipped as a release by mistake. `setup-prefix` takes about 2m30s reading the winetricks cache straight out of the read-only store. Launching `ableton-live` creates the runtime link, registers the GC root, and writes the handler entries through it. Live's Hardware Setup button opens the PipeASIO panel, confirmed by hand on the VM.

Fedora, against a tarball runtime: prefix creation from scratch and `--refresh` both work, `[6/6]` does the right thing with and without `ABLETON_PREFIX_MANAGED`, and no runtime link gets created. `[6/6]` also installed Live 12 Suite from a real 6.6 GB zip with nothing left behind. A `.run` on the same VM covered update (2026.08.14.3 to 2026.08.19.1), uninstall and fresh install.

All seven suites pass with no failures: `test-nix-packaging` (29), `test-pipeasio-installer` (86), `test-installer-lifecycle` (72), `test-desktop-integration` (3), `test-shortcut-hold` (45), plus `test-tsan-policy` and `test-release-policy`. `make check` passes too.

**Known issue**

`setup-link.sh` does not work from a packaged install. `enable_link()` starts with `[ -x "$linkctl" ]`, and `linkctl` resolves only to `$ABLETON_DATA_HOME/ableton-linkctl`, which nothing populates on Nix, so `.#setup-link` fails with advice a Nix user cannot act on. Lines 1159 and 1232 in the same file already fall back to `$here/ableton-linkctl`, which does resolve here. It is pre-existing on `main`, so I have left it out of this PR; 

**Follow-ups**

`[6/6]` reimplements the payload handling in `installer.sh:753-800`, because the package deliberately doesn't ship that script. Moving that path into `scripts/lib/` so both callers share it would remove the duplication properly, and `newest_live_zip` should move with it.

Nothing in CI builds the flake, and `flake.nix` has no `checks` output, so `nix flake check` only evaluates.

Left out to keep this to packaging: the `niri` probe, the busctl/dbus-send theme fallbacks, the `check-ntsync.sh` robustness work, and the README section. The first three are worth landing on their own.


Immabed:
Flake seems to be working decently, but I'm seeing an issue with the PipeASIO settings (when clicking "Hardware Setup" in Audio settings). On my Mint machine the PipeASIO settings panel opens, but on nix I get the following pop-up:
image
Additionally, I cannot launch `pipeasio-settings` in a terminal, it is not a recognized command. I can modify the settings in `~/.config/pipeasio/config.ini` and that does update PipeASIO.

ClickSentinel:
@Immabed Thanks for finding that. Fixed with latest push. 
image

shibco:
waow gonna look at this tonite @ClickSentinel

flake.nix
```diff
@@ -0,0 +1,80 @@
...
+        meta.description = "Print the read-only audio diagnostic snapshot an issue report is expected to carry";
+      };
+      setup-realtime = {
+        type = "app";
+        program = "${self.packages.${system}.ableton-wine}/share/ableton-wine/scripts/setup-realtime.sh";
+        meta.description = "Install the distribution-canon pro-audio profile (rtprio, swappiness, governor; needs sudo)";
+      };
+      setup-link = {
+        type = "app";
+        program = "${self.packages.${system}.ableton-wine}/share/ableton-wine/scripts/setup-link.sh";
```
shibco, line 75: The flake publicly exports `setup-link` here but the package places `ableton-linkctl` beside the script, and `setup-link.sh` only searches `$ABLETON_DATA_HOME`. This means the enable action will immediately fail the executable check.

The derivation substitutes exact runtime and daemon paths, and `setup-link` writes those into the user config and systemd unit. For what it's worth, I think that contradicts the PR’s own `stable-link/GC-root` rationale

nix/ableton-wine.nix
shibco, line 1: Installing the PipeASIO icon first creates `.../scalable/apps

But the later `cp -a .../scalable .../scalable` on L385-388 will produce: `share/icons/hicolor/scalable/scalable/apps/live-suite.svg`

This is a broken path and we therefore need validation.

scripts/build-audit.sh
```diff
@@ -85,9 +85,50 @@ esac
...
+pass=0; failed=0; skipped=0
 ok()  { pass=$((pass+1));   printf '   %-42s PASS %s\n' "$1" "$2"; }
 bad() { failed=$((failed+1)); printf '   %-42s FAIL %s\n' "$1" "$2"; }
+# ABLETON_AUDIT_PROFILE=nix relaxes ONLY the container-pipeline provenance
+# records (sanitizer runs, builder manifests, installer helper hashes, the git
+# source-tree digest): the Nix build runs none of those pipelines and cannot
+# stamp records it has not produced. Every structural, patch, fingerprint and
+# binary-hash check still fails hard, and the release profile is unchanged.
+# A relaxed record is reported and counted as skipped, never as passed.
+AUDIT_PROFILE="${ABLETON_AUDIT_PROFILE:-release}"
```
shibco, line 97: This audit accepts any all-/nix/store/ `RPATH` regardless of` ABLETON_AUDIT_PROFILE`. Consequently, the normal release audit can approve a tarball binary that only runs with a particular Nix closure.

shibco (changes requested):
Our build pipeline doesn't test the flake, and omit `flake.nix`, `flake.lock`, and `nix/**`.

@ClickSentinel - can we add those triggers and run both nix flake check and an actual nix build .#ableton-wine?

There are only a few small tweaks needed before we can merge this, imo. Good work everyone <3

Immabed:
Sample log file running commit 3d36184eec2e4146daed1fc30733bf96cf0f1768 on NixOS.
[live.log](https://github.com/user-attachments/files/31353797/live.log)

shibco:
GOATed, thanks @Immabed. I'll take another look after the rest of the review is done!

ClickSentinel:
image

shibco:
haha omg wow we just merged nix support

PR #240, opened 2026-08-22
Fix ClearType bitmap strikes

shibco:
Use glyph outlines for ClearType and retain bitmap strikes for fonts that depend on them.


shibco:
Actually, i'll rebase this into one commit

shibco:
Thanks to @giang17 and @ClickSentinel for providing a lot of research for this cleartype implementation.

@giang17 - this might be of use to your repo.

giang17:
Thanks for picking this up, Cade!

One clarification for the record: this is the fix we offered back on #155 — our commit [`21a37233c`](https://github.com/giang17/wine/commit/21a37233cbc2ac3aaa8ffc495f2e5bdccfb81650), written while running your nine ClearType patches here, to close the embedded-strike gap in 0081 for Tahoma / MS Shell Dlg. So the direction on this one was the other way round — there is nothing for us to take back. Happy to see it in the series, and the `SERIES.sha256` / `build-audit.sh` handling is clean.

One follow-up of ours that this port does not include, in case it is useful: Windows selects strike vs outline by **rendering mode** (GDI-compatible modes take the strike, NATURAL modes take the outline), not by whether ClearType was asked for. Without that, greyscale text still takes the strike, and since Tahoma does not carry a strike for every size, neighbouring sizes flip between hard 1-bpp bitmaps and antialiased text at 96 dpi. Ours is opt-in behind `HKCU\Software\Wine\DirectWrite` "outline_in_natural_modes" — [`2a00f9b`](https://github.com/giang17/wine/commit/2a00f9b7a51a7779db85e5b7fe24b633cdfa73fe). It extends the same `freetype_load_glyph` helper, so it should sit on top of your 0100 without conflicts. Both applications we instrumented request NATURAL exclusively (Studio Pro 8: 1625 of 1625 glyph-run analyses), so the strike is being taken in exactly the modes where Windows would not take it.

ClickSentinel:
The rule is keyed on whether a ClearType texture was requested, which fixes ClearType and leaves the greyscale path taking bitmap strikes.

`create_glyphrunanalysis()` selects `DWRITE_TEXTURE_ALIASED_1x1` whenever the antialias mode is `GRAYSCALE`, whatever the rendering mode, so `lcd` is FALSE there and the outline is never requested. Direct2D selects that mode whenever `can_draw_cleartype()` fails: a non-opaque target, FLAT pixel geometry, or an explicit greyscale request.

Measured on Wine's Tahoma at 11 ppem, inside its strike range, greyscale antialias mode: 1 distinct coverage value with 14 samples at exactly 255, which is one-bit output. ClearType at the same size and face gives 44. So the two select different forms for one face at one size, and both differ from GDI.

`dlls/win32u/freetype.c` has the rule already: it sets `FT_LOAD_NO_BITMAP` for every requested format except `GGO_BITMAP`. Deriving the flag from the rendering mode rather than from `lcd` covers greyscale too, and it removes a size-to-size inconsistency, since Tahoma carries strikes at 13 and 15 but not 14 and neighbouring sizes otherwise render in different forms.

That is more than a one-liner, since the chosen form has to join the glyph cache key and travel in both unixlib structures. I have it built and measured if it saves you the work: `ClickSentinel:fix/cleartype-outline-strikes-amended`, one commit on top of this branch. After it, greyscale at 11 ppem gives 13 to 15 distinct values with at most 1 saturated, ClearType above the strike range is unchanged, and ALIASED and the GDI modes stay byte-identical to an unpatched runtime.

shibco:
The rendering-mode feedback is now actioned in the stacked #242 branch. Its reviewed 0100 uses outlines only for NATURAL and NATURAL_SYMMETRIC, while GDI_CLASSIC, GDI_NATURAL, and ALIASED retain bitmap strikes. The runtime probe now covers both natural modes plus all three strike-preserving modes. This base branch was also merged with current main so the stack retains the newer installer and launcher work.

shibco:
Final review pass: current main is merged, and c58f1fe moves the corrected rendering-mode rule into this PR itself. NATURAL and NATURAL_SYMMETRIC use outlines; GDI_CLASSIC, GDI_NATURAL, and ALIASED retain bitmap strikes. The Wine series is 90 patches through 0102. Exact-head CI passed: https://github.com/shibco/ableton-linux/actions/runs/32663781037

PR #242, opened 2026-08-22
Preserve fractional glyph positions

shibco:
Use 16 horizontal positions per pixel for natural rendering and store each rendered glyph by position.


ClickSentinel:
0101 adds `subpixel_phase` to the glyph cache key, multiplying its space by 16, while `fontface_cache_init()` stays at `max_size = 0x8000`. A 54-glyph run at em 12 then no longer fits three phases, so every glyph is re-rasterised on every draw.

Measured, 200 iterations, ClearType 3x1:

    em 12  NATURAL      0.2246 ms/run     GDI_NATURAL  0.0078 ms/run
    em 24  NATURAL      0.3760 ms/run     GDI_NATURAL  0.0131 ms/run

GDI_NATURAL is the control: same binary, same texture type, same outline path, phase 0. A phase sweep puts the cliff between 2 and 3 phases and flat after it, and a 6-glyph run stays at 0.0024 ms, so it is a capacity miss rather than per-phase rasterisation cost.

One line fixes it:

    fontface->cache.max_size = 0x8000 * DWRITE_SUBPIXEL_PHASES;

That takes the em 12 comparison to 0.0128 against 0.0075. The residual is the real cost of rasterising 16 distinct phases instead of reusing one entry.

Two smaller notes. The narrowing to `NATURAL` and `NATURAL_SYMMETRIC` is right and I verified it holds: ALIASED and both GDI modes hash identically to an unpatched runtime. And a glyph that loads as a strike still gets a phase computed and its box widened while `freetype_get_aa_glyph_bitmap()`'s bitmap branch never applies the translate, so it gains a dead column and 16 identical cache entries. That disappears once 0100 stops handing strikes to the natural modes, so it needs no separate fix.

Branch if you want the commit: `ClickSentinel:fix/dwrite-subpixel-positioning-amended`.

shibco:
Addressed the cache-capacity feedback by integrating #252, including scaling the per-face budget by DWRITE_SUBPIXEL_PHASES and running the glyph gate in CI. I also corrected the amendment's overly broad non-ALIASED form rule: only NATURAL and NATURAL_SYMMETRIC use outlines; both GDI modes and ALIASED retain strikes. The expanded probe now performs 12 checks, and the branch is based on current main. Local results: all 99 Wine patches through 0101 apply cleanly; make check and the complete host suite pass.

shibco:
Final review pass: the corrected #240 parent is merged here, and 0103 keeps 16 fractional phases while scaling the glyph cache by the same factor. The 12-case runtime gate passed, including natural outline/subpixel behavior and GDI/aliased strike/whole-pixel behavior. Exact-head CI passed: https://github.com/shibco/ableton-linux/actions/runs/32663782804

PR #243, opened 2026-08-22
Preserve D2D target format on resize

shibco:
Reuse the target pixel format and bitmap settings so resize keeps ClearType text rendering.


ClickSentinel:
The mechanism is right and I could reproduce the defect it fixes: with an hwnd render target created `D2D1_ALPHA_MODE_IGNORE`, per-channel spread over drawn text is 67.44 before a resize and 0.00 after it without this patch, and 64.79 either side with it.

The explicit bitmap options introduce a second problem though. Naming `D2D1_BITMAP_OPTIONS_TARGET | D2D1_BITMAP_OPTIONS_CANNOT_DRAW` drops `D2D1_BITMAP_OPTIONS_GDI_COMPATIBLE`. A target created with `D2D1_RENDER_TARGET_USAGE_GDI_COMPATIBLE` gets a swapchain flagged `DXGI_SWAP_CHAIN_FLAG_GDI_COMPATIBLE` (`hwnd_render_target.c:913`), so its buffer carries `D3D11_RESOURCE_MISC_GDI_COMPATIBLE` and `d2d_get_bitmap_options_for_surface()` reports the option for it.

Nothing rejects the narrower request. `d2d_bitmap_check_options_with_surface()` only fails when the caller asks for GDI_COMPATIBLE and the surface lacks it, never the reverse, so the bitmap is simply built without the flag and nothing fails until `GetDC`. Measured with a GDI-compatible target: GetDC succeeds before the resize and returns 0x8899001a, `D2DERR_TARGET_NOT_GDI_COMPATIBLE`, after it. Without this patch it succeeds either side.

One line, and it is what passing no properties did anyway, so it preserves `CPU_READ` and anything added later rather than special-casing one flag:

    bitmap_desc.bitmapOptions = d2d_get_bitmap_options_for_surface((IDXGISurface *)dxgi_surface);

taken after `GetBuffer`, with the pixel format assignment left where it is. Branch if useful: `ClickSentinel:fix/d2d-cleartype-resize-amended`.

shibco:
Addressed the resize feedback in a3279e3. After GetBuffer, bitmapOptions now comes from d2d_get_bitmap_options_for_surface(), preserving GDI_COMPATIBLE (and future surface-derived flags) while retaining the target pixel format. Added a runtime probe for both ClearType survival and GetDC survival across Resize, and wired it into CI beside the glyph gate. Both probes compile with -Werror; make check and the complete host suite pass locally.

shibco:
Follow-up from the first combined CI run: the MinGW headers do not expose ID2D1Bitmap1_Map/Unmap as COBJMACROS, so the probe now uses the interface vtable directly and compiles cleanly with -Wall -Wextra -Werror. The failure path also revealed an unbounded cleanup call against an unset prefix; 2e577b5 assigns the throwaway prefix before compilation, stops Wine before Xvfb, and bounds prefix setup, probe execution, and cleanup. I cancelled the stuck run and pushed this fix for a clean rebuild.

shibco:
Final review pass: the current #242 stack is merged here, and 0104 preserves the target pixel format while deriving bitmap options from the resized DXGI surface. All 12 DirectWrite checks plus ClearType-across-Resize and GetDC-across-Resize passed. Exact-head CI passed: https://github.com/shibco/ableton-linux/actions/runs/32663788319

PR #253, opened 2026-08-23
dwrite: key the strike-or-outline rule on the rendering mode

ClickSentinel:
0100 requests the outline when the caller wants a ClearType texture. That fixes ClearType and leaves the greyscale path taking bitmap strikes, so the two select different forms for one face at one size, and both differ from GDI.

`create_glyphrunanalysis()` selects `DWRITE_TEXTURE_ALIASED_1x1` whenever the antialias mode is `GRAYSCALE`, whatever the rendering mode, so `lcd` is FALSE there and the outline is never requested. Direct2D selects that mode whenever `can_draw_cleartype()` fails, which covers a non-opaque target, FLAT pixel geometry, and an explicit greyscale request.

`dlls/win32u/freetype.c` sets `FT_LOAD_NO_BITMAP` for every requested format except `GGO_BITMAP`. Deriving the flag the same way gives greyscale, ClearType and GDI the same form for one face at one size, and it removes a size-to-size inconsistency: Tahoma carries strikes at 13 and 15 but not 14, so neighbouring sizes otherwise render in different forms.

Measured on Wine's Tahoma at 11 ppem with the greyscale antialias mode: before, 1 distinct coverage value with 14 samples at exactly 255, which is one-bit output; after, 13 to 15 distinct values depending on the sub-pixel phase, with at most 1 saturated. ClearType above the strike range is unaffected, and ALIASED, GDI_CLASSIC and GDI_NATURAL stay byte-identical to an unpatched runtime (FNV-1a over the alpha texture: `ddb8d79287340624` and `9d19ccee0c9bd876`).

The chosen form joins the glyph cache key, since a strike and an outline of one glyph at one size differ in content and cannot share an entry. It is carried in both unixlib structures so the bounding box and the rasterised form cannot disagree about which was used, which means both halves of dwrite must be rebuilt and deployed together.

### DIFF
--- a/patches/0100-dwrite-use-outlines-for-cleartype-glyphs.patch
+++ b/patches/0100-dwrite-use-outlines-for-cleartype-glyphs.patch
@@ -1,84 +1,153 @@
-From: Giang Nguyen <33104857+giang17@users.noreply.github.com>
-Date: Fri, 14 Aug 2026 10:11:28 +0000
-Subject: dwrite: use outlines for ClearType glyphs
+Subject: dwrite: choose strike or outline by rendering mode
 
-Some fonts include small, pre-drawn glyph images called bitmap strikes.
-Each strike provides one brightness value for each pixel. ClearType needs
-3 horizontal samples for each pixel.
+An embedded bitmap strike carries one coverage sample per pixel, which is the
+aliased form. Antialiased and ClearType output need more than one sample per
+pixel, so a strike supplied for either request is expanded back into aliased
+coverage. Wine's own Tahoma, the face MS Shell Dlg resolves to and so the
+default interface font, carries strikes for 8 to 13, 15 and 16 ppem, which
+covers most interface text.
 
-Wine uses the glyph outline for ClearType text. Wine retains bitmap strikes
-when they are the font's sole glyph source. Wine uses the same glyph form to
-calculate the boundary and draw the pixels. Wine stores each result. Later
-requests use the stored result.
----
+dlls/win32u/freetype.c sets FT_LOAD_NO_BITMAP for every requested format except
+GGO_BITMAP: the strike is the aliased form, so any other requested format
+requires the outline. Derive the flag from the rendering mode rather than from
+whether a ClearType texture was requested. That gives greyscale, ClearType and
+GDI the same form for one face at one size, and it removes a size-to-size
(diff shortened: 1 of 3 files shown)


shibco:
The new explicit bitmap options at `0102-d2d1-keep-target-pixel-format-on-resize.patch#L25-L26` omit `D2D1_BITMAP_OPTIONS_GDI_COMPATIBLE`. Consequently, a GDI-compatible `HWND` render target becomes non-compatible after a resize, returning `D2DERR_TARGET_NOT_GDI_COMPATIBLE`. 

To fix this, we need to explicitly preserve the flag when `render_target->desc.usage` requests GDI compatibility. That flag is specifically what enables surface `GetDC`. 

I confirmed this against[ Microsoft documentation](https://learn.microsoft.com/en-us/windows/win32/api/dxgi/ne-dxgi-dxgi_swap_chain_flag)

Additionally, for this PR and #252 , the new glyph “gate” isn't covered by CI. It is instead a standalone Make target,`dwrite-subpixel-positioning/.github/workflows/ci-pr-build.yml:132`. Probably this makes a future glyph regression pass the build check.

shibco (changes requested):
Blocking on the rendering-mode boundary at cb22bc6. The current rule uses rendering_mode != DWRITE_RENDERING_MODE1_ALIASED, which also forces outlines for GDI_CLASSIC and GDI_NATURAL. Those are GDI-compatible modes; the upstream follow-up keeps bitmap strikes for both GDI modes and uses outlines only for NATURAL and NATURAL_SYMMETRIC. The green build does not exercise that semantic distinction. I integrated the corrected helper and expanded runtime coverage into #242 at 822489f/9220d39: GDI_CLASSIC, GDI_NATURAL, and ALIASED retain strikes, while both natural modes use outlines. Please do not merge the current #253 head; it will be superseded when the corrected stack lands.

shibco:
Superseded by the corrected #240/#242 stack. #240 now keys outline selection strictly on NATURAL and NATURAL_SYMMETRIC, preserving GDI_CLASSIC, GDI_NATURAL, and ALIASED bitmap strikes; current main also owns patch numbers 0100 and 0101. This amendment would reintroduce the reviewed GDI regression and stale numbering, so it should not be merged.

PR #259, opened 2026-08-24
Draft: ARM64 Architecture Support

sajattack:
PipeASIO is not working.

Build Audit is disabled.

I tried vendoring the pipeasio from x86_64 because it doesn't build as aarch64 and I thought maybe FEX could handle the x86_64 version, but the regsvr command fails so I should probably walk that back once again.

Several pre-existing patches in the build-audit do not apply to aarch64; when I changed the build audit to detect the native architecture, I got 50 or so failures.

For some reason I have to `rm -rf drive_c/Users/user/AppData/Roaming/Ableton` every time I launch it.

Also need to remove `drive_c/ProgramData/Ableton/Live 12/Resources/Core Library/Lessons/Demo Songs` or it crashes trying to load the demo.

Ableton Index is constantly crashing and bringing up annoying popups.
Screenshot_20260823_235453


shibco:
Surely there's an ARM64 version of PipeASIO?
If not, that would be a good think to ask that project nicely for.

PR #260, opened 2026-08-24
tiling WM: disable cursor warp and document focus mitigations

midisaur:
Running Live under a tiling WM (Hyprland, i3, sway, river, etc.) reveals a focus issue that standard desktop environments mask: Live uses `SetCursorPos` internally (automation editing, clip nudging), and on tiling compositors this forces the cursor to the centre of the window, triggering workspace-switching breakage and an apparent focus steal.

**What this PR does:**

1. **`setup-prefix.sh`** — sets `MouseWarpOverride = disable` in the Wine X11 Driver registry key. This tells Wine to ignore `SetCursorPos` calls from the application. Safe here because Live never depends on absolute cursor position for critical functionality. Pairs naturally with the existing `FileDialogPortal` key right above it.

2. **`notes/ABLETON-TILING-WM-FOCUS.md`** — a reference note explaining the symptom and both sides of the fix: the Wine registry key and the optional compositor-side `focus_on_activate = false` for Hyprland/niri/river (Wayland compositors that respect the property).

Tested on Hyprland 0.55.4 (PikaOS 4 / Debian trixie), Live 12 Suite, project Wine 11.13 stack.

### DIFF
--- a/notes/ABLETON-TILING-WM-FOCUS.md
+++ b/notes/ABLETON-TILING-WM-FOCUS.md
@@ -1,0 +1,55 @@
+# Tiling WM / tiled layout — focus and cursor behaviour
+
+## Background
+
+Ableton Live's `SetCursorPos` calls (used during automation editing, clip
+nudging, and other UI gestures) are harmless on conventional desktop
+environments. On **tiling compositors** (Hyprland, i3, sway, river, dwm, etc.)
+they manifest as an apparent focus-steal: the cursor jumps to the centre of
+Live's window, the compositor tracks the activation event that triggered the
+warp, and the user is yanked back to Live's workspace when they were trying to
+switch away.
+
+## Registry fix — `MouseWarpOverride = disable`
+
+Applied at prefix setup (see `setup-prefix.sh` step 5/6):
+
+```
+wine reg add 'HKCU\Software\Wine\X11 Driver' \
+  /v MouseWarpOverride /t REG_SZ /d disable /f
+```
+
+This tells Wine to ignore `SetCursorPos` calls from the application. Live
+never depends on absolute cursor position for critical functionality, so the
+override is safe.
+
+## Compositor-side: `focus_on_activate = false`
+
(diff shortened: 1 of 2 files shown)


shibco (changes requested):
Hey @midisaur! Thanks for looking into this, and for testing it on Hyprland! I also saw your repo and am excited to work with you on VST support!!

We can't merge this yet! I took a look tonight and here's what I found:

I checked the registry change against the Wine source we build from, and it doesn't do anything. Wine only reads `MouseWarpOverride` in `dlls/dinput/mouse.c`, and it reads it from `HKCU\Software\Wine\DirectInput` in the registry. Nothing reads it under X11 Driver. Even under the right key, the value only stops DirectInput's own centre warp for games that grab the mouse. It doesn't touch `SetCursorPos`. Live doesn't use DirectInput at all; it calls `SetCursorPos` from `user32` directly. So the line in `setup-prefix.sh` doesn't change any behaviour, on any desktop.

To highlight something additionally important: Live also _does_ depend on those warps - this is what we use to  moves the pointer during fader and knob drags, and our pointer patch exists to repair those drags when the warps don't land on XWayland. So we couldn't ship a switch that ignores `SetCursorPos` even if Wine had one; it would bring back the drifting faders.

However, the Hyprland setting in this PR is the real fix. Hyprland itself defaults `misc:focus_on_activate` to `false`, but PikaOS's default `hyprland.lua` **turns it back on**, and Omarchy and CachyOS do the same. When it's on, Hyprland focuses the window on an activation request and then warps the cursor into it. I am almost certain that's the jump to the centre of Live's window you saw. 

Could you retest with the registry key removed and only `focus_on_activate = false` set? 

If my suspicions are correct, youll get the same result.

A few final things from the local LLM pass on this:

- niri and river don't have a focus_on_activate option. sway and i3 have focus_on_window_activation. Please drop niri, river and dwm.
- `#` isn't a comment in Lua, so the snippet won't load; use `--`. Please also add the `hyprland.conf` form for people who haven't moved to the Lua config.

Thank you! Looking forward to merging!

PR #261, opened 2026-08-24
nix: stage every scripts/lib helper, not a list that goes stale

ClickSentinel:
`nix run github:shibco/ableton-linux` exits before it starts:

    ableton-live: live-options.sh helper is missing or incomplete

scripts/lib/live-options.sh predates the flake landing on main. The launcher sources it and refuses to run without three of its functions, and install.sh and make-installer.sh both stage it, but nix/ableton-wine.nix named four helpers in a list and it was not among them. The packaged launcher therefore cannot start at all.

The installCheckPhase gate did not catch it because it restated the same four names: the staging and the check meant to cover the staging went stale together. Both now derive from scripts/lib itself, so a helper added later is covered without editing this file, and the gate also asserts the three functions the launcher tests for rather than that a file exists - a truncated helper passes a presence check and still produces the reported message.

This is the third gap of this shape, after audio-report.sh and the icon paths, each one a hardcoded list here drifting from what the tarball ships. Hence the glob rather than a fifth name.

tsan.sh is staged as a result. install.sh does not stage it and nothing in the package reads it; it costs a few hundred bytes against not having to judge, per file, whether the launcher will come to need it.

Verified on nixos-cloud from this branch: flake check passes, audit 171 checks, all six helpers staged. `ableton-live` with no arguments reaches "Wine prefix is missing or incomplete", the correct next failure, where main stops at the helper message. Note --version does NOT show this: it answers at line 158, before the check at line 263, on both trees.

### DIFF
--- a/nix/ableton-wine.nix
+++ b/nix/ableton-wine.nix
@@ -165,9 +165,13 @@ stdenv.mkDerivation {
         # The launcher and max9 resolve their config/lifecycle helpers beside
         # themselves first ($launcher_here/lib), then under the user's
         # ~/.local/share/ableton-wine (which only installer.sh populates).
+        # Every lib, by glob, not a list: the launcher hard-fails on a helper it
+        # cannot source, and a list here goes stale the moment one is added
+        # upstream. live-options.sh arrived that way and stopped the launcher
+        # starting at all, with the gate below naming the same stale four.
         mkdir -p $out/libexec/lib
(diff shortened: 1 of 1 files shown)


shibco:
Works on my machine!

BlueskyFR:
Hey there! What is the intended way to setup & launch Ableton on NixOS using this project?
Should we run the installer or do `nix run` every time?

PR #264, opened 2026-08-24
Small fix for some edge cases with hotplugging 

shibco:


ClickSentinel (approved):
no blockers, only recommendation is not publishing the data ports to other wine processes.


```diff
-    port_out = snd_seq_create_simple_port(midi_seq, dest->token != 1 ? port_out_name : "WINE ALSA Output",
-                                          SND_SEQ_PORT_CAP_READ | SND_SEQ_PORT_CAP_SUBS_READ | SND_SEQ_PORT_CAP_SUBS_WRITE,
+    port_out = snd_seq_create_simple_port(midi_seq, dest->token != 1 ? port_out_name : "WINE ALSA Output",
+                                          SND_SEQ_PORT_CAP_READ | SND_SEQ_PORT_CAP_SUBS_READ |
+                                          SND_SEQ_PORT_CAP_SUBS_WRITE | SND_SEQ_PORT_CAP_NO_EXPORT,
```

Other comments:
WINEDEBUG=+midi faults the new reservation path. midiwatch.exe SIGSEGVs after ASSERT READY_FOR_READD under a 15 s exclusive reservation: 3 of 3 with +midi or err+midi, 0 of 1 with -all; the plain replug case under +midi is clean on both runtimes, so this is not patch 0028's generic "any MIDI trace faults" shape. The pre-fix runtime fails the busy case before reaching that code, so new-with-264 against newly-reachable is unsettled. Users are safe (the launcher sets -all); the people it hits are the ones debugging this feature. Fix: no code change proposed, file it.

The probe gives a released reservation a flat 2 s. run_cycle_case sets test_timeout_ms = block_ms + 2000, so a 15 s block fails and the same block with 40 s of slack passes — a timing artefact, not a recovery failure. It is inert at the shipped 3000 ms. Fix: derive the allowance from the documented 10 s retry period rather than a constant.

PipeASIO's location gate only bites when both records carry the field. Compiled src/hotplug.c standalone: saved bus path pci-0-usb-1 against a candidate reporting none scores 2 by node name, 1 by device name; against the moved twin with both paths present, 0. So the twin protection depends on PipeWire reporting a bus path for the returning node. 

Fix: say so in the note's matching rules, or reject when the saved record has a location field the candidate lacks.
Two CHANGELOG entries are not user-visible. The ALSA event pool size and "the rapid reconnection test completes its removal step after it converts Windows line endings" describe internals and a test fixture; "we fixed" is first person against the rest of the file. Fix: cut both entries and reword the lead as a symptom.

PR #269, opened 2026-08-26
[Moonshot PR] CPU optimiser pipeasio hotpath

shibco:


ClickSentinel:
**No blockers.** A delivered callback leaves every output slot empty, so the publish 0013 removes was doing no work: 4096 of 4096 slots already clear over 2048 graph cycles, on a build carrying a temporary counter. Measured through pipeasio's headless ASIO hosts against a runtime built at `08a9e129`, including a bit-exact `tests/asio_loopback` sweep, 4 of 4 phases at 128/256/512/1024 frames, twice here and twice on the installed pre-0013 runtime; nothing is owed in Live, the question is a per-cycle counter no Set reports.

Other comments:

`pipeasio_pw_output_needs_fallback` is `return !delivered`, in a header otherwise about `pw_buffer` regions, and the first new test group guards a `pipeasio_pw_finish_output` call with `if (pipeasio_pw_output_needs_fallback(true))`, which never runs. Fix: test the condition directly and drop the helper with its two assertions; the two `finish_output` groups keep their value.

```diff
--- a/include/pipeasio_pw_buffer.h
+++ b/include/pipeasio_pw_buffer.h
@@
-/* After Live returns audio, the callback owns each output and queues it once.
- * During recovery, the backend owns each output and sends one silent
- * block. */
-static inline bool
-pipeasio_pw_output_needs_fallback(bool delivered)
-{
-    return !delivered;
-}
-
 static inline bool
 pipeasio_pw_finish_output(_Atomic(struct pw_buffer *) *cycle_slot, const float *source,
--- a/src/audio.c
+++ b/src/audio.c
@@
-            if (pipeasio_pw_output_needs_fallback(delivered))
+            if (!delivered)
```

"During a buffer-size change" is one of the three conditions that leave a slot full; `process_callback` returns `admitted ? 0 : 1`, so Stop and teardown reach the fallback too. Fix:

```diff
--- a/src/audio.c
+++ b/src/audio.c
@@
-                /* During a buffer-size change, the backend owns the PipeWire
-                 * output. It sends one silent block at the current PipeWire
-                 * buffer size. After Live returns audio, the callback owns the
-                 * output and queues it once.
-                 * Each audio block has one owner. */
+                /* A delivered cycle has already exchanged and queued every
+                 * output slot.  The three cases that leave one full publish
+                 * here instead, one silent block at the current PipeWire
+                 * buffer size: no process callback installed, a callback the
+                 * host gate rejected (Stop or teardown), and a graph quantum
+                 * that disagrees with the host buffer size. */
```

`include/audio.h` calls the nonzero return "the recovery result", where recovery already names the reconnection worker in `src/audio.c` across 54 occurrences, and the rewrite drops the old "(no buffers, or stopped)". Fix:

```diff
--- a/include/audio.h
+++ b/include/audio.h
@@
-/* Zero means the ASIO host completed the cycle and queued each output once.
- * The recovery result gives each output to the backend for one silent block.
- * The backend also uses that result to tell a deliberate Stop from
- * an xrun. */
+/* Returns 0 when the ASIO host consumed the cycle and queued each output once,
+ * nonzero when it did not (no buffers, or stopped), which leaves every output
+ * to the backend for one silent block.  The backend also uses this to tell a
+ * real xrun apart from the idle cycles of a deliberate Stop. */
```

The note gives 750 calls/s without the per-call cost. The removed call is one acquire load on `route_ready[AUDIO_PLAYBACK_LATENCY]` plus one acq_rel exchange on an empty slot, 5.48 ns compiled standalone at `-O2` over 200M iterations, and `audio_on_process` fires per graph cycle rather than per ASIO buffer: at `default.clock.quantum = 1024` and `PIPEASIO_DEFAULT_OUTPUTS` of 2 that is 94 calls/s, about 0.5 us/s. Fix: give the per-call cost beside the rate.

Nothing in the build exercises `audio_on_process`. `pipeasio_ctest_nonintegration` runs `ctest -LE '^integration$'` and `tests/asio_loopback` carries `LABELS "integration;wine;pipewire"`, so "deterministic unit coverage for both ownership paths" is coverage of `pipeasio_pw_finish_output` in two argument configurations. Fix: name the loopback analyzer as the test that covers the ownership decision.

PR #271, opened 2026-08-26
[Moonshot PR] CPU optimiser bench reporting tool

shibco:


ClickSentinel:
No blockers. Two things the report asserts are not true of the run it describes: the worker count is always the pre-launch value, and a set that stopped playing mid-window was reported clean. Exercised on a full 5-set run in Live 12.4.3 with `--skip-vst-check`, plus the tool's own pipeline headless, both new suites, `shellcheck` and `--dry-run`.

**Should fix**

- The report always states the pre-launch worker count. `bench-suite.sh:196` runs `profile` before the first launch and the launcher writes `-MaxAudioThreads` on launch: profile captured at `14:25:27.737Z`, `Options.txt` written at `14:25:32.314Z`, so `report.md` printed `MaxAudioThreads=Live default` for a session that ran with 10. `live_install` then compares two identically stale records, so a change in worker count between runs reads as a match. Fix: re-sample `live_inventory(prefix)` after the first `wait_for_owned_live`.
- A set that stops playing mid-window is reported clean. `Benchmark_Max4Live` recorded `/abl/bench/playing 0` at offset 5.590833 of its 30 s window, and its DSP load fell from 7.410% over the first 5 s to 1.332% over the last 5 s. `parse_osc` counts the address into `addresses` at `bench-report.py:714` and nothing reads it. Cause unsettled at n=1; the arrangement-loop flag reads true in all 4 controlled sets. Fix: confound a playback-mode set that records any `/abl/bench/playing` during the window.
- `probe`'s nonce is typed as a number about 1 call in 760, so 1 suite run in 190. `encode()` types any argument parsing as an int or a float as one; over 4,000,000 `token_hex(8)` values, 0.053% raise `struct.error` or `OverflowError` and 0.078% encode as a float, where the pong never matches and `probe` waits out the 180 s `--launch-timeout`. Fix:
  ```diff
  --- a/scripts/bench-osc.py
  +++ b/scripts/bench-osc.py
  @@ -118,7 +118,7 @@ def probe(timeout: float) -> int:
       Send a random value until the matching reply arrives.
       """
       deadline = time.monotonic() + timeout
  -    nonce = secrets.token_hex(8)
  +    nonce = "n" + secrets.token_hex(8)
       next_ping = 0.0
       with socket.socket(socket.AF_INET, socket.SOCK_DGRAM) as listener:
           listener.bind(("127.0.0.1", RECV_PORT))
  ```
- A value absent from both runs counts as a run detail difference, in 6 of the 30 `IDENTITY_SPECS`. `identity_comparison` returns `missing-both` with `confounder: True` whenever `allow_empty=False` and neither side has a value, so with `powerprofilesctl` absent a run compared against itself reports "Run detail differences: 1". `gpu`, `gpu_renderer`, `audio_profile`, `sample_rate` and `quantum` carry the same flag. Fix:
  ```diff
  --- a/scripts/bench-report.py
  +++ b/scripts/bench-report.py
  @@ -2605,7 +2605,7 @@ IDENTITY_SPECS = (
       ("configuration", "Launcher and PipeASIO settings hash", ("configuration", "identity_hash"), False),
       ("pipeasio_config", "PipeASIO settings path and hash", ("configuration", "pipeasio_effective_config"), False),
       ("benchmark_gates", "Wine, PipeASIO, and launcher CPU settings", ("configuration", "benchmark_gates"), True),
  -    ("power_profile", "Power profile before Live starts", ("system", "power_profile"), False),
  +    ("power_profile", "Power profile before Live starts", ("system", "power_profile"), True),
   )
  ```

Other comments:

- The osc collector is rejected on start latency alone, on 2 of 4 controlled sets. `collector_command_usable` applies `first_output <= min(1.0, duration * 0.1)` to `osc` as well as `pw-top`: Empty 1.8825 s and Max4Live 1.1877 s against Inbuilts 0.0983 s, yet all four collected 56 or 57 samples. Fix: give `osc` an edge allowance covering the device's 500 ms report period.
- `task-churn` fires on 5 of 5 sets. `snapshots_delta` sets `churn = bool(born_keys or exited_keys or thread_born_keys or thread_exited_keys)`, and each set had 1 to 2 threads born and 1 to 2 exited against 177 to 182 survivors. It reaches every comparison as `before:task-churn` and `after:task-churn`. Fix: report the counts and threshold the boolean.
- `--duration` under 10 seconds turns crackle instrumentation off for every set: `pw-top -b` emits one batch a second and the edge tolerance falls below that gap. Unusable at 3, 5, 8 and 9, usable at 10, 11 and 15; `positive_duration` accepts 1 to 3600. Inert at the shipped 30. Fix: reject a duration under 10, or name the floor in `bench/README.md`.
- `bench/results.csv` and `bench/workload.csv` are referenced by nothing in the tree, and appear in neither `bench/SHA256SUMS` nor any `.gitignore`. Fix: drop both, or name the command that reads them.
- `markdown_report` prints `pending` for three states: command not installed (`:2408`), file absent (`:2445`), value not measured. Fix: render the first two as `not installed` and `absent`.
- Two Live installs at one version render as identical rows. `:2443` labels each record by `version` alone, so `users/lgillingham/.../Live 12.4.3` and `users/Public/.../Live 12.4.3` both print `Live 12.4.3`. Fix: include the `users/<name>` segment.

PR #272, opened 2026-08-26
[Moonshot PR] CPU optimiser hybrid topology

shibco:


ClickSentinel:
No blockers. Six suites pass; measured through the report's own function against this box's sysfs (12th Gen i5-12600KF, 6 P-cores plus 4 E-cores, kernel 7.0.11), the six new fixtures, and synthetic zero-CPU roots. Base Wine read at `base-5c23dd1c` `dlls/ntdll/unix/system.c`.

Should fix:

- `scripts/lib/live-options.sh:162` declares `cpus` with no assignment, so `[ "${#cpus[@]}" -gt 0 ]` at `:174` is an unbound-variable abort under `set -u` whenever `$cpu_root/cpu[0-9]*` matches nothing. `rc=1`, `cpus: unbound variable`; the same call with `cpus=()` returns `rc=0` and the full degraded block. Not reachable on a host with `/sys` mounted, but it makes the function's own zero-CPU branch (`allowed_source=unavailable`, `smt_evidence=unknown`) dead code that no fixture can reach.
  ```diff
  -    local -a cpu_dirs cpus rows=()
  +    local -a cpu_dirs=() cpus=() rows=()
  ```
  ```diff
  -    printf '%s\n' "${rows[@]}"
  +    [ "${#rows[@]}" -eq 0 ] || printf '%s\n' "${rows[@]}"
  ```
  Fix: both hunks, plus a seventh fixture with no `cpuN` directories.
- Gate step 12 fires on every Intel hybrid, so the gate cannot authorise the rule step 1 tests for. On this rig `core_type` does not exist (nor does `/sys/devices/system/cpu/types/`) and `cpu_capacity` is 1024 on all 16 threads, P and E alike; only `cpu_core/cpus`=`0-11` against `cpu_atom/cpus`=`12-15` separates them. `core_type=unavailable` is partial evidence, so step 12 keeps Linux placement even where the P/E split is unambiguous. Fix: score class evidence on any one sufficient field rather than on all five, and drop `wine_win32_efficiency_class_probe` from the class list, since its constant `not_run_read_only` is neither `unavailable` nor `invalid` and the rule has no outcome for it.
- `heterogeneous_evidence` has no third state, unlike `smt_evidence` (`unknown`) and `preferred_core_evidence` (`unavailable`). A fixture with readable topology and zero readable class fields prints `heterogeneous_evidence=not_observed topology_fields_complete=yes`, identical to a symmetric machine with full evidence; `topology_complete` is cleared only by `physical_package_id`, `core_id` and `thread_siblings_list`, never by `cpu_capacity` or `core_type`. On this box it reads `yes` while all 16 rows carry `core_type=unavailable`. Fix: add `heterogeneous_evidence=unknown` when no class field is readable, or a `class_fields_complete` line so step 12 is decidable from the summary.

Other comments:

- The section is 5425 of 12713 bytes, 43% of the whole audio report, at 16 threads: one ~250-character row per allowed CPU. The AMD dev host's 32 threads reach ~60%, a 128-thread host ~43 KB, in the file whose stated job is to be pasted into an issue. Runtime is 0.26 s, so the cost is size. Fix: emit the per-CPU rows only when an evidence line reads `present`.
- Two of three CHANGELOG sub-bullets record no change: "Linux keeps control of CPU placement" was already true, and "future automatic placement requires matched audio tests" is a plan. The entry also drops the bold user-facing symptom lead every other release block uses. Fix: cut both, lead with what a user gains, for example "**Audio reports now carry CPU layout evidence**". Nothing parses `CHANGELOG.md`, so the new `## Unreleased` heading is safe.
- `wine_topology_online_cpus` names processors Wine does not enumerate above 64 threads. `system.c:1182` is `if (i > 8 * sizeof(ULONG_PTR)) break;`, and `:1160` reads `/sys/devices/system/cpu/present` for its `MAXIMUM_PROCESSORS` FIXME, which the launcher hides with `WINEDEBUG=-all`. Fix: record `present` beside `possible`, and note the index-64 stop next to the field.
- The note's Wine account holds: `:1110` opens `/sys/devices/cpu_core/cpus`, `:1168` opens `/sys/devices/system/cpu/online`, `:897` derives `EfficiencyClass` from the `performance_cores` bitmap, and no patch in the 94-patch series touches that file. No action.
- The base moved: 269 and 271 merged since `5493354`. `git merge --no-commit --no-ff upstream/main` conflicts in `CHANGELOG.md` alone; `Makefile` and `TROUBLESHOOTING.md` auto-merge. No action beyond the rebase.

Coverage: `test-cpu-topology.sh` 6/6, `test-desktop-integration.sh` 6/6 (after staging `dist/pipewire-version-probe`, `ableton-linkd`, `cabextract-static` and the 2026.08.26.1 tarball, without which it aborts before its assertions), `test-live-options.sh` 30/30, `test-nix-packaging.sh` 30/30, `test-release-policy.sh` 6/6. `bash -n` and `shellcheck -S warning` clean on all four changed scripts.

PR #273, opened 2026-08-26
[Moonshot PR] CPU optimiser pipeasio telemetry

shibco:


ClickSentinel:
One blocker: `pipeasio/0013` is already taken on main, so the series cannot merge as numbered. Measured through a minimal ASIO host against the built runtime (audit 179/179, 106 patches), plus the four telemetry cycle functions extracted verbatim into a standalone harness.

**Blocking**

- `pipeasio/0013-callback-phase-telemetry.patch` collides with main's `pipeasio/0013-avoid-redundant-output-fallback-publish.patch`, merged in `4b320a9` after this branch was cut at `5493354`. `git merge --no-ff upstream/main` conflicts in `patches/SERIES.sha256` and `scripts/build-audit.sh`, whose `REQUIRED_PIPEASIO_TAIL` already names that 0013; PR 274 claims the number as well. Fix: renumber to 0014/0015, repoint the tail constant, and correct the notes doc's "apply after PipeASIO patch 0012".

**Should fix**

- 0014's reopened `/proc/self/fd/2` carries its own file offset, so ordinary `PIPEASIO_LOG` writes overwrite telemetry records whenever stderr is a regular file. 3 s run with `PIPEASIO_DEBUG=1`: 13 of 22 telemetry lines survive to a file, 22 of 22 through a pipe. `scripts/ableton-live:404` makes stderr a pipe, so the invocation the notes give is clean; a manual `2> file` capture is not. `tests/unit/test_telemetry.c` never writes through the original descriptor after the reopen, so it cannot see this.
  ```diff
  +/* fd 2 as a regular file cannot block the worker, and a second open file
  + * description carries its own offset: O_APPEND writes telemetry at EOF while
  + * fd 2 keeps writing behind it. Reopen only for the pipe and tty case. */
  +static int
  +telemetry_open_report_fd(void)
  +{
  +    struct stat st;
  +    if (fstat(STDERR_FILENO, &st) != 0)
  +        return -1;
  +    if (S_ISFIFO(st.st_mode) || S_ISCHR(st.st_mode))
  +        return open("/proc/self/fd/2", PIPEASIO_TELEMETRY_OUTPUT_FLAGS);
  +    return dup(STDERR_FILENO);
  +}
  ```
  ```diff
  -                else if ((state->output_fd = open("/proc/self/fd/2",
  -                                                  PIPEASIO_TELEMETRY_OUTPUT_FLAGS)) < 0)
  +                else if ((state->output_fd = telemetry_open_report_fd()) < 0)
  ```
  Fix: both hunks in 0014's `src/asio.c`, helper beside `telemetry_log`. All 3 records then survive in order on a file, and the pipe arm still returns EAGAIN on backpressure; `sys/stat.h` is already included at `src/asio.c:42`.
- `notes/MOONSHOT-CPU-PIPEASIO-TELEMETRY.md` says each measured callback reads a clock 6 times; it is 8, because 3 phases need 4 boundaries at 2 clocks each. `strace -c -e trace=clock_gettime` on the built runtime gives 565 syscalls gate-on against 0 gate-off over 141 cycles, so 4.0 `CLOCK_THREAD_CPUTIME_ID` reads per cycle plus 4 vDSO `CLOCK_MONOTONIC`. Fix: 8, and add that 4 of them are syscalls at 137.5 ns against 12.1 ns for `CLOCK_MONOTONIC`, 598 ns per cycle.

**Comment**

- A muted cycle spends 6 of the 8 clock reads and publishes nothing, and `telemetry_emit` writes no line at all when `cycles` and `dropped` are both 0, so a run held in issue 49's fail-silent mismatch mode looks exactly like telemetry being off. Sustained mute is unmeasured here: 2 hosts at 1024 and 256 frames each negotiated their own quantum. Fix: count muted cycles into the `cycles=` record.
- No action: the off switch is exact-match, unlike #121's. `off`, `0`, `disabled` and `'off '` each give 0 telemetry lines and no warning; `-1` and `bogus` give 0 lines and the warning. Unset gives 0 `clock_gettime` syscalls and no output.

PR #274, opened 2026-08-26
[Moonshot PR] CPU optimiser pipewire arbitration

shibco:


ClickSentinel:
Blocking: `pipeasio.node` is absent from a node's registry global, so `pipeasio_node_signature` is never true and 0013 admits no PipeASIO peer at all. Measured through two concurrent headless ASIO hosts and a libpipewire registry-global probe, against this head's build (audit 178/178, 105 patches) and the pre-274 runtime.

**Blocking**

- `pipeasio_node_signature` requires `marker == "1"`, and the registry global for a filter node has 7 props with `pipeasio.node` not among them: `object.serial`, `client.id`, `node.description`, `node.name`, `media.type`, `media.category`, `media.role`. The property is set (`src/audio.c:928`) and `pw-dump` shows it on the same node's `info.props`; only the global omits it, and the global is the dict `audio_cache_node` receives. Two same-name hosts at 256 and 128: pre-274 caches the peer (id 106) as ours, this head caches only its own node, one `pipeasio-quantum-converge` WARN either way so the peer was live in both. `foreign_forcers` stays 0 on both, and the same-name peer that used to sit in the cache is now dropped.
  ```diff
  -    return marker && media_type && media_category && media_role && !strcmp(marker, "1")
  -           && !strcmp(media_type, "Audio") && !strcmp(media_category, "Duplex")
  +    return media_type && media_category && media_role && (!marker || !strcmp(marker, "1"))
  +           && !strcmp(media_type, "Audio") && !strcmp(media_category, "Duplex")
              && !strcmp(media_role, "DSP");
  ```
  Fix: that line in `include/pipeasio_parse.h`; `tests/unit/test_config.c`'s `EXPECT_TRUE(!pipeasio_node_signature(NULL, "Audio", "Duplex", "DSP"))` flips with it. Rebuilt with it folded in (audit 178/178): peer second gives `node 107 foreign force-quantum enabled`, peer first gives `reclassified node id=113 as foreign after own id=107`. With no PipeASIO running, 0 of 9 registry nodes match the three `media.*` keys.

Other comments:

- `audio_reclassify_nodes`'s `PIPEASIO_NODE_IGNORE` branch cannot execute, so "Remove rejected provisional nodes" does not describe the built driver. `IGNORE` turns only on `!signature && !audio_class`; both are written once at cache time (`src/audio.c:3196`, `:3202`) by the same test that already gated admission, and `audio_on_node_info` refreshes neither. Enumerating the classifier's whole input space: 16 cacheable combinations, 0 reclassifying to `IGNORE`, 8 flipping ours to foreign. No arm on any runtime printed `rejected provisional node`. Fix: drop the branch and its `audio_remove_cached_node` call, or refresh `media_class` and `node_name` from the props event if catching a late rename was the intent.
- `notes/PIPEASIO-15.md`'s new paragraph states behaviour the build does not have: "PipeASIO assigns each application's buffer request to that application" and "PipeASIO treats Live or PipeASIO sessions as separate when they share a name" are both false at this head, and neither names the bound node id the change rests on. Fix: rewrite after the signature fix, naming the id as the discriminator.
- PRs 269, 273 and 274 each add a `patches/pipeasio/0013-*.patch` and each re-pin `REQUIRED_PIPEASIO_TAIL` to their own; 273 adds an `0014` as well. Fix: no code change, sequence the merges and renumber the two that follow.
- No action: registration is complete. `SERIES.sha256`, the `pipeasio/0013|ascii|…|pipeasio-node-arbitration` fingerprint and the tail pin are all present, `patch-check.sh` passes at 105 manifest entries, and `ABLETON-WINE-PATCH-STACK.txt` in the shipped tarball carries `f1b87ece…`.

PR #275, opened 2026-08-26
[Moonshot PR] CPU optimiser ntsync proof

shibco:


ClickSentinel:
`make test` is red at this head: `test-desktop-integration.sh` greps two strings `check-ntsync.sh` no longer prints, and its fake runtime has no `bin/wineserver` for the new gate. Measured through the shipped suites plus `check-ntsync.sh` against `wine-d2d1-nspa-11.13` on a host with `/dev/ntsync`, and the launcher block driven through a `uname -r` shim.

**Blocking**

- `test-desktop-integration.sh:173` greps `FAIL: ntsync not compiled in`; the checker now prints `FAIL: rebuild Wine with linux/ntsync.h`. Its fixture at line 164 stages only `bin/wine`, so the new `check-ntsync.sh:58` wineserver gate exits 1 first and the static grep is never reached. CI exit 2; locally exit 1 at 2 of 6, and 6/6 with the diff below. Line 171's `! grep -q 'semantics probe not found'` is vacuous for the same rename.
  ```diff
   printf '#!/bin/sh\nexit 0\n' > "$base/fake-wine/bin/wine"
  -chmod +x "$base/fake-wine/bin/wine"
  +printf '#!/bin/sh\nexit 0\n' > "$base/fake-wine/bin/wineserver"
  +chmod +x "$base/fake-wine/bin/wine" "$base/fake-wine/bin/wineserver"
  @@
  -! grep -q 'semantics probe not found' "$base/ntsync.err" \
  +! grep -q 'Provide the NTSync probe at' "$base/ntsync.err" \
       || fail "installed NTSync diagnostic cannot find its packaged sibling probe"
  -grep -q 'FAIL: ntsync not compiled in' "$base/ntsync.err" \
  +grep -q 'FAIL: rebuild Wine with linux/ntsync.h' "$base/ntsync.err" \
  ```
  Fix: stage the wineserver beside the wine stub, update both greps.

**Should fix**

- `scripts/lib/manifest.sh:814` recognises an installed `check-ntsync.sh` by `NT sync semantics hold`, which the rewrite deleted. `ableton_legacy_owned_path` sourced standalone: main's copy OWNED, this head's NOT OWNED, a foreign script under the same name NOT OWNED. Swapping to a new string inverts it, so the pre-manifest copy needs the old one kept.
  ```diff
  -        "$ABLETON_DATA_HOME/check-ntsync.sh") grep -qF 'NT sync semantics hold' "$path" 2>/dev/null ;;
  +        "$ABLETON_DATA_HOME/check-ntsync.sh")
  +            grep -qF 'NT sync semantics hold' "$path" 2>/dev/null \
  +                || grep -qF 'ABLETON_REQUIRE_NTSYNC' "$path" 2>/dev/null ;;
  ```
  Fix: OR the two markers.
- `TROUBLESHOOTING.md:181` quotes the old success line and line 184 says any other final line means NTSync is inactive, so a working setup now reads as a failure.
  ```diff
  -OK: sync semantics hold, ntsync active (server holds /dev/ntsync)
  +OK: NTSync active; sync semantics pass; wineserver opens /dev/ntsync
  ```
  Fix: that line, and add the exit 3 case for a host without the device.

Other comments:

- `ABLETON_NTSYNC_WARN` is exact-match `off`, so `OFF`, `0`, `no` and `false` all still warn. That matches `ABLETON_RT` at `scripts/ableton-live:1343`, so the launcher is consistent; the note is the odd one, telling users `ABLETON_REQUIRE_NTSYNC=0` for the sibling switch. Fix: use `off` at `notes/MOONSHOT-CPU-NTSYNC-PROOF.md:18`.
- `flake.nix:56` still describes the app as verifying "NT sync semantics hold" and does not say a device-less host now exits 3 without running the probe. Fix: reword the description.
- Mechanism holds: 27/27 probe assertions and `server_dev_ntsync_fds=1`, cleanup leaving no `/tmp/check-ntsync.*` and sparing seven unrelated wineservers. The fd count reads 1 on all seven, the `Live 12*` Options.txt pattern matches both real files in `~/.wine-ableton`, and the kernel arithmetic is right on nine strings including `6.14-rc1` and a bare `6`. No action.

ClickSentinel:
All five findings from the previous pass are fixed and CI is green (`ci-pr-build` 17m03s, `ci-flake` 37m30s). `check-ntsync.sh`, `ableton-live`, `audio-report.sh`, `test-check-ntsync.sh` and the `Makefile` are byte-identical to `6e5da50`, so this pass re-ran the earlier harnesses against the moved consumers: `test-desktop-integration.sh` goes from exit 1 at 2 of 6 to exit 0 at 7/7, and the removed-string sweep that found four stale consumers now finds none.

Other comments:

- `test-desktop-integration.sh:172` sources `$here/lib/manifest.sh`, the source-tree copy, while every neighbouring assertion reads the installed artefact under `$base/data/ableton-wine/`. `install.sh:758` stages `lib/manifest.sh` into `$data/lib/`, so a stale or unstaged installed copy passes this gate. Repointed, the suite still gives 7/7.
  ```diff
  -        legacy-ntsync "$here/lib/manifest.sh" "$ownership_check"
  +        legacy-ntsync "$base/data/ableton-wine/lib/manifest.sh" "$ownership_check"
  ```
  Fix: read the installed copy.
- `TROUBLESHOOTING.md:186` reads "Exit status 1 or 2 means that the NTSync proof failed". Exit 2 is a usage error on both its paths: `ABLETON_REQUIRE_NTSYNC=invalid` gives `!! Use on or off for ABLETON_REQUIRE_NTSYNC`, and `ABLETON_TEST_NTSYNC_DEVICE` outside the hermetic suite gives `!! ABLETON_TEST_NTSYNC_DEVICE is reserved`. Neither ran a proof, so the sentence sends a user with a typo to the issue tracker.
  ```diff
  -Exit status 1 or 2 means that the NTSync proof failed.
  +Exit status 1 means that the NTSync proof failed. Exit status 2 means that a
  +setting passed to the check was not valid.
  ```
  Fix: split the two statuses.
- The new seventh assertion covers all three marker arms and cannot pass vacuously: sourcing `manifest.sh` alone needs no `config.sh`, and an undefined function returns 127, which the two positive arms catch before the negative one runs. No action.

PR #276, opened 2026-08-26
[Moonshot PR] CPU optimiser pipewire pro audio

shibco:


ClickSentinel:
No blockers. One recommendation: allow a pro-audio-first pair, which the tool currently refuses. Exercised end to end against an idle HDMI card (device 48) on PipeWire 1.5.85 and WirePlumber 0.5.12, plus the suite's 20 fake-graph cases in 12 s.

- `already-pro-audio` (`scripts/pipewire-pro-audio-ab.sh:885`) refuses a device already on pro-audio, so every pair is baseline first, pro-audio second (`:936`, `:1019`). Boost residency, page cache and thermal drift all load onto the B leg. Fix: accept a pro-audio start and record the order in `status.json`.
- A stale lock blocks every later run for that device. Lock staged by hand: exit 73, "Device 51 already has an active Pro Audio comparison", path in neither the message (`:838`) nor `status.json`; the retry with the same `--output` then exits 2, because the failed run already populated the directory. `notes/` covers SIGKILL and power loss with `recovery-command.txt`, which is the same case. Fix: name the lock path in the message and in that section.
- `blockers.active_links` records the peer by id alone (`:368`). Device 49 refused on `output_node_id: 89`, which is Firefox; node 89 carries no `device.id`, so no name reaches the artifact and "Close each PipeWire client for the target device" cannot be acted on from the report. Fix: carry the peer's `node.name` and `application.name` into the blocker record.
- The one comparison example calls `scripts/bench-suite.sh` (`notes/PIPEWIRE-PRO-AUDIO-AB.md:95`), which is on `moonshot-cpu-optimiser-bench-report` and on no other branch; line 87 asks for "an integration worktree that contains both scripts" and names none. Fix: name that branch, or use `scripts/bench-run.sh`, which is on main.
- `notes/PIPEWIRE-PRO-AUDIO-AB.md:139` dates a session and cites device 69, an id PipeWire reassigns across restarts. Fix: cut the line, keep the USB-interface caveat as a plain limit.
- `--dry-run` is a silent alias for `--discover` (`:31`, `:63`) yet it records state, creates the report directory and can exit 73. Fix: drop the alias.
- The transient switch leaves nothing behind. Device 48 round trip: 6/6 `restored_evidence` verified, and independently `default-profile`, `default-routes`, `default-nodes` and the device's `Profile` and `Route` params byte-identical either side. B really ran on pro-audio, one `hdmi-stereo` node against 4 `pro-output-*` with `AUX0..AUX7`. `save:false` is honoured at `state-profile.lua:77`, and routes take the opposite default at `find-best-routes.lua:52`, so `default-routes` was worth checking separately. No action.

```text
Next, on the real interface: scripts/pipewire-pro-audio-ab.sh --device 49 --wine-prefix ~/.wine-ableton --output /tmp/ab-49 -- /bin/true, with every PipeWire client closed first. Watching for: .restored_evidence.defaults and .routes in status.json, which an idle non-default card cannot exercise.
```

ClickSentinel:
One blocker: the Pro-Audio-first path this commit adds cannot run on WirePlumber 0.5. Re-ran the five harnesses from the first pass across both revisions on PipeWire 1.5.85 and WirePlumber 0.5.12; suite 23/23 in 10 s, `shellcheck -S warning` clean on both scripts.

**Blocking**

- `stored_profile_name` (`scripts/pipewire-pro-audio-ab.sh:533`) reads `pw-metadata -n default-profile`, an object WirePlumber 0.5.12 does not create: `pw-metadata -l` lists 8 and that is not one of them, and the only `default-profile` in `/usr/share/wireplumber` is `State ("default-profile")` at `scripts/device/state-profile.lua:130`, which is the state file. `before/default-profiles.canonical.txt` is therefore 0 bytes on every run. The function returns empty against that real file and `output:analog-stereo+input:analog-stereo` against a synthetic metadata line, so the parser is right and the source is empty; the answer it wants is on disk for device 49 in `~/.local/state/wireplumber/default-profile`, the file `capture_profile_state_hash:366` already reads. End to end, device 48 on pro-audio: `baseline-profile-missing`, exit 65, `order: []`. The suite passes because the fake emits a `default-profile` object at `scripts/test-pipewire-pro-audio-ab.sh:166`, which models 0.4. Fix: parse the state file, and drop the `default-profile` case from the fake so the suite models 0.5.

Other comments:

- `.legs` gains a third key. Any failure before the profile branch emits `B_comparison`, measured on `concurrent-run` and `baseline-profile-missing`, while `target-device-active` and `discovery-ok` emit `B_pro_audio`. `notes/` documents only `B-pro-audio` and `B-baseline`. Fix: emit the resolved key, or document `B_comparison`.
- The other 5 findings are fixed and re-measured. Lock: message and `.lock_path` both name `/run/user/1000/ableton-pw-pro-audio-1000-51.lock`, exit 73, staged by hand. Peer: a real client on device 49 now records `peer: {id: 129, node_name: "pw-cat", application_name: "pw-cat"}` against a bare `output_node_id: 89` at `a3add3b`. `--dry-run` exits 2. The note names the branch that carries `bench-suite.sh` and drops the dated line. No action.
- The phase refactor leaves the restore path intact. Device 48 forward pair: `ab-complete`, `order: ["baseline","pro-audio"]`, 6/6 `restored_evidence` verified, and independently `default-profile`, `default-routes`, `default-nodes` and the device's `Profile` and `Route` params byte-identical either side. No action.

```text
Next, on the real interface: scripts/pipewire-pro-audio-ab.sh --device 49 --wine-prefix ~/.wine-ableton --output /tmp/ab-49 -- /bin/true, with every PipeWire client closed first. Watching for: .restored_evidence.defaults and .routes, which an idle non-default card cannot exercise. Device 49 is also the only card here with a saved regular profile, so once the source is the state file it is the one that can run the pair in reverse.
```

PR #278, opened 2026-08-26
[Moonshot PR] CPU optimiser live workers

shibco:


ClickSentinel:
No blockers. The automatic count rises on every SMT machine below 16 physical cores, and the 16-core computer the notes cite is the one shape where the new floor is inert. Measured through `test-live-options.sh` (32/32), `test-desktop-integration.sh` (6/6), and arms against `scripts/lib/live-options.sh` and the launcher's `auto` case block cut by pattern.

Next, needs Live on an SMT box below 16 physical cores, after merging `upstream/main` for `scripts/bench-suite.sh`:

```text
env ABLETON_MAX_AUDIO_THREADS=10  scripts/bench-suite.sh --tag phys
env ABLETON_MAX_AUDIO_THREADS=15  scripts/bench-suite.sh --tag floor
env ABLETON_MAX_AUDIO_THREADS=off scripts/bench-suite.sh --tag live-own
```

Values are a 10-core, 16-thread i5-12600KF. Watching for: `AudioCalc` worker count and deadline load on the Max4Live and VSTs sets.

**Should fix**

- `reliability_floor` at `scripts/lib/live-options.sh:85` halves `live_default`, which is derived from `nproc`, so the floor is `logical - 1` below 17 threads and 16 at or above. It tracks the thread count, not the core count.

  | machine | phys | logical | live own | main auto | PR auto |
  |---|---|---|---|---|---|
  | 4c/8t laptop | 4 | 8 | 14 | 4 | 7 |
  | i5-12600KF, 6 P-cores plus 4 E-cores | 10 | 16 | 30 | 10 | 15 |
  | 16c/32t, the cited computer | 16 | 32 | 31 | 16 | 16 |

  On the cited computer the floor is `(31 + 1) / 2 = 16` and the physical count is also 16, so `max()` returns 16 under either policy and that measurement cannot separate them. Fix: derive the floor from the physical count, or run the three arms above before merge.
- `ableton_reliable_audio_threads` is absent from the helper list at `scripts/ableton-live:263`, so a stale `$ABLETON_DATA_HOME/lib/live-options.sh` passes the completeness check and `auto` degrades to `off` plus an audio-report message, not the refusal `scripts/test-desktop-integration.sh:118` asserts.
  ```diff
   for live_options_function in ableton_available_physical_cores \
  +                             ableton_reliable_audio_threads \
                                ableton_live_product_version \
                                ableton_seed_max_audio_threads; do
  ```
  Fix: that line, nothing else.

Other comments:

- `CHANGELOG.md:14` and `TROUBLESHOOTING.md:525` carry "Before release, test busy Live Sets on computers with 4 to 8 physical cores", an engineering task in two user-facing files. The range understates it: every SMT machine below 16 physical cores changes. Fix: cut both, keep `notes/DEFAULT-ON-FEATURE-AUDIT-2026-08-23.md:102` and widen its range.
- "A number from one to 63 requests that limit" in `notes/FINDINGS-PIPEASIO-CPU-2026-08-20.md` is still false at this head, on the sentence the PR rewrites. 32-thread host: 30 writes, 31 and 63 write nothing and say nothing. 8-thread host: 13 writes, 14 writes nothing. Fix: state that a number at or above Live's own count is ignored.
- 64 or more physical cores fails the range check at `live-options.sh:76`, so `auto` takes the fallback and prints "Run .../audio-report.sh to review the CPU details" against a topology that read fine. 63 clean, 64 and 96 print it. The chosen value is still correct, and still better than main's `received '64'`. Fix: accept any positive count, since the cap at `live_default` already bounds it, and drop 64 from the `for invalid` loop in `test-live-options.sh`.
  ```diff
  -    case "$physical" in [1-9]|[1-5][0-9]|6[0-3]) ;; *) return 1 ;; esac
  +    case "$physical" in ''|*[!0-9]*) return 1 ;; esac
  ```
- `ableton_live_calculated_audio_threads` replaces two hand-copied copies of the same arithmetic in `ableton_seed_max_audio_threads`, and the 1-to-32 loop in `test-live-options.sh` covers both branches. No action.
- `upstream/main` moved to `f224fb8` under the PR. The merge is clean, and both suites still pass on it. No action.

PR #279, opened 2026-08-26
More hardening of the installer

shibco:


ClickSentinel:
One blocker: `ableton_backup_launcher` writes `<name>.bak` outside both recovery mechanisms, so a successful install destroys a user's own `.bak` and cannot put it back. Measured through both installer suites (lifecycle 77/77, PipeASIO 89/89, rc 0) and three XDG-sandbox harnesses; `bash -n` and `shellcheck -S warning -x` clean, merge with `upstream/main` at `f224fb8` clean.

**Blocking**

- `manifest.sh:1063` gives `<launcher>.bak` only `ableton_txn_snapshot`, which is per-run, while every other displaced object gets `ableton_persist_file_prestate`. A user's own `~/.local/bin/ableton-live.bak`: successful install overwrites it, nothing under `state/` holds it, uninstall does not restore it. The arm where `update-desktop-database` fails does restore it, so only the success path loses the file. Fix: make the backup a managed path.
  ```diff
  +    ableton_persist_file_prestate "$backup" "" replace-launcher || return 1
       ableton_txn_snapshot "$backup" || return 1
       ableton_txn_expect "$backup" "$current" || return 1
  -    ableton_atomic_restore_object "$target" "$backup"
  +    ableton_atomic_restore_object "$target" "$backup" || return 1
  +    ableton_record_owned "$backup" "${current%%:*}"
  ```
  Needs `*.bak) ableton_launcher_path_allowed "${path%.bak}" && return 0 ;;` in `ableton_managed_path_allowed`. That tree gives PipeASIO 89/89 and lifecycle 71/72, the one failure being `test-installer-lifecycle.sh:1775`, which asserts the backup survives uninstall. Those two cannot both hold: `remove_owned_manifest_files` walks the manifest, and an unmanaged path has no prestate slot to restore a user file from.

Should fix:

- CHANGELOG calls `.bak` "each displaced file or symlink"; it holds the most recently displaced object. Install over a foreign `ableton-live.desktop` puts the foreign entry there, then replaying the launcher's own `Name=` and `StartupWMClass=` heal and updating leaves `.bak` holding the project's entry. Prestate keeps the original, so recovery is intact. Fix: reword to the most recently replaced object.
- Uninstall leaves all 5 backups it wrote and names none of them, so each restored original sits beside a byte-identical copy. `test-installer-lifecycle.sh:1775` pins this deliberately, so the gap is that nothing tells the user. Fix: state in the CHANGELOG that `<name>.bak` outlives uninstall and is the user's to delete.

Other comments:

- `desktop_entry_is_foreign` is deleted with no note in the body, which is empty, or the CHANGELOG. `wine-protocol-c74max.desktop` is the name winemenubuilder writes for a registered protocol handler (`programs/winemenubuilder/winemenubuilder.c:2184`), so the installer now displaces another package's entry and pins `x-scheme-handler/c74max` to its own. Fix: name the reversal in the CHANGELOG entry.
- `mkfifo ~/.local/bin/ableton-live` then `--integration-only` exits 1 with only `!! component installation failed`, 0 mentions of the path on stderr. Base `54933547` hangs instead, `timeout 25` returning 124. Fix: an `ableton_config_error` beside the directory guard in `ableton_backup_launcher`.

PR #283, opened 2026-08-28
New Installer

shibco:
Rebuilt the installer again, this time from code that was set aside but more tested and more permissive/tolerant to unexpected encounters in its working directory.

@ClickSentinel - there is one more commit coming to add settings options, but that shouldn't mess with your review. pls review ASAP thx


ClickSentinel:
Two blockers: the packer cannot pack, and install never updates the ownership manifest, so uninstall keeps the files it just wrote and reports them as user-modified. Measured through the nine touched suites (293 assertions), a packed dev `.run` driven through update and uninstall on a live install, and a fresh-install sandbox against `ff4a27f`.

**Blocking**

- `make-installer.sh:152` copies `scripts/lib/ui.sh` into the kit; the allowlist at `verify-installer-payload.py:130` still omits it, so `[4/5] wrapper self-check` fails with `!! installer payload contains unexpected member: scripts/lib/ui.sh` and no `.run` packs at this head. CI run 33182474247, step 14. With the token added, the verifier returns 0 against a real packed `.run`.
  ```diff
  -    for name in ("config.sh", "lifecycle.sh", "live-options.sh", "manifest.sh", "pipeasio.sh"):
  +    for name in ("config.sh", "lifecycle.sh", "live-options.sh", "manifest.sh", "pipeasio.sh", "ui.sh"):
  ```
  Fix: that one token.
- `ableton_install_project_file` (`lib/manifest.sh:1587`) never calls `ableton_record_owned`, so `install-manifest.tsv` is never rewritten. On an upgrade the manifest keeps the previous install's digests, `uninstall.sh:1058` finds every changed file mismatched and takes the `else` at `:1090`. Live install: manifest last written `2026-08-23 16:52`, update ran `08-28 13:27` and did not touch it, uninstall at 13:39 kept `VERSION`, `audio-report.sh`, `rollback.sh`, `setup-link.sh` and `lib/{config,lifecycle,manifest,pipeasio}.sh` as `kept an unrecognised or user-modified file`, advising the user to move them aside. All 7 are byte-identical to the kit that had just written them; `lib/ui.sh` has no row and went silently. Footer read 16 warnings, **Complete**. Fresh install is the same cause with no warning at all: `ff4a27f` writes 45 rows and leaves 0 files, `108a0ae` writes none and leaves 18. Fix: record ownership in `ableton_install_project_file` and `ableton_install_project_symlink` and publish the manifest as the install completes.

Other comments:

- Uninstall asks whether to overwrite files. `setup-link.sh:647` saves the off mode through `ableton_install_project_file`, so a removal run reaches `QUESTION: Some files from an earlier installation already exist`, default `[O]verwrite all`, `Existing files will be moved to .../backups/... before they are replaced`, five-second timeout. Fix: write the mode file directly on the disable path, or skip the question when `ABLETON_UI_ACTION` is `uninstall`.
- A running spinner draws through the question block. Same uninstall, step 3/3: `Stop Ableton Linux background services ⣷` overwrote the option lines and interleaved with the prompt and the timeout hint. `test-installer-ui.sh` never runs `ui_question` beside a `ui_run`, so none of its 19 assertions can see it. Fix: hold the spinner while a question owns the trunk, and add the concurrent case.
- `test-installer-ui.sh` aborts at assertion 3 wherever `awk` is mawk, so 17 of 19 never run and `make test` fails with it. The simulator at `:58` moves the cursor with `length`/`substr`, and `LC_ALL=C.UTF-8 mawk '{print length($0)}'` on `│` prints 3, so `ESC[nC` lands inside the box-drawing bytes and `│  │` renders as `││` on 13 lines. The same `live.raw` through a codepoint-correct simulator reproduces the golden byte for byte and the suite passes 19/19. mawk is the Debian, Ubuntu and Pop!\_OS default; the GitHub runner carries gawk, so CI stays green. Fix: decode to codepoints in `vt.awk`, or gate on a multibyte awk the way the suite already gates on `script(1)`.
- No CHANGELOG heading for the version being released. `VERSION` is `2026.08.28.1` and the PR commits `dist/BUILD-INFO-2026.08.28.1.txt` and its `.sha256`, while `CHANGELOG.md:3` reads `## Unreleased`. `test-release-policy.sh` passes at 9/9: nothing in `scripts/` or the `Makefile` mentions `CHANGELOG`. Fix: retitle the block `## 2026.08.28.1`.

Edited after testing

ClickSentinel:
Both blockers are fixed. Two findings survive, both in the uninstall path, both now with their own harness. Measured through the touched suites at this head, an upgrade sandbox that seeds a pre-283 manifest before installing over it, and a two-process pty fixture for the renderer.

- `verify-installer-payload.py:130` carries `"ui.sh"`. A `-dev.run` repacked at this head verifies rc=0 with no local patch. Fixed, no action.
- Uninstall no longer strands its own support files. `harness/upgrade-leftovers.sh` installs with `ff4a27f`'s tree so a manifest exists with the old digests, upgrades with the PR's tree, then uninstalls: `108a0ae` gave 6 `unrecognised or user-modified` warnings and 7 files left, `abedf95` gives 0 and an empty tree. Fresh install went 18 files left to none. Fixed, no action.
- The manifest is still never written, so the fix rests entirely on `authoritative_support_relatives` at `uninstall.sh:80-91` staying in step with what `install.sh:800-812` publishes. Two fixed lists, in two files, with nothing comparing them. `test-uninstall-boundary.sh` pins the 22 current entries, not the correspondence. Fix: derive one list from the other, or add an assertion that every path `install.sh` writes under `$ABLETON_DATA_HOME` appears in the inventory.

Other comments:

- The overwrite question fires during `uninstall`. `setup-link.sh` and `lib/manifest.sh` are byte-identical to `108a0ae`, so the call at `setup-link.sh:647` stands. Real `setup-link.sh disable` under a pty against a seeded config: rc 0, `Some files from an earlier installation already exist` on screen with the timeout hint. The sandbox uninstall arms miss it because Link is never enabled there. Fix: write the mode file directly on the disable path, or skip the question when `ABLETON_UI_ACTION` is `uninstall`.
- Two processes draw the tree with no arbitration. `ui.sh`'s change is colour scoping; `test-installer-ui.sh:250` still asserts it uses neither `tput` nor `flock`. A child that inherited `ABLETON_UI_TTY_FD` raising `ui_question` while the parent's `ui_run` spins: the parent drew **133** spinner frames after the child posted the prompt, and its settled `├─ Check the embedded kit ✓` landed below the timeout hint, inside the question block. Same fixture in one process renders clean, which is the control. That is the topology of the live symptom, `setup-link.sh` prompting under uninstall's own spinner. Fix: have the child signal the parent to hold the spinner while a question owns the trunk, and add the two-process case.
- Deleting INSTALLER.md's "What you see" took the only user-facing record of the backup path with it. `grep -rn 'ableton-wine/backups' *.md` is now empty and only a CHANGELOG bullet mentions the prompt; `TROUBLESHOOTING.md` has nothing. The tree description was fair to cut, but a five-second default that moves a user's existing files, and the directory it moves them to, are what a user needs written down. Fix: restore those two paragraphs, prompt and log location, without the tree walkthrough.
- `make-installer.sh:236` writes `-dev.run`; `:257` prints the release name in the command it hands you. Fix: `$(basename "$out")` at `:257`.

Suites at this head: `test-installer-ui` 22/22 as shipped, up from 2/19 under mawk; `test-release-policy` 12/12; `test-link-boundary` 20/20; `test-launcher-transactions` 27/27; `test-desktop-integration` 12/12; `test-project-file-installer` 9/9. `test-uninstall-boundary`, `test-installer-lifecycle` and `test-pipeasio-installer` are still running and are not yet part of this verdict.

ClickSentinel:
Four blockers. The first must not ship: the new Wine gate is an unscoped copy of a mechanism this repo already has, it kills every Wine process on the machine, and accepting defaults cancels the install. Measured through the twelve touched suites on an idle box, a three-arm stub `.run`, and install and uninstall runs on a real Ableton install.

**Blocking**

- `setup-run-header.sh:692-698` re-implements, unscoped, what `install.sh` already does correctly. `pgrep -x wineserver` matches any wineserver on the machine (8 on this box, all from unrelated test prefixes); the default is `[N]o - exit the installer`, taken after 5 s; yes runs `pkill -x wineserver` across all of them. Driving the menu with Enter picks `[I]nstall`, defaults through the six preflight questions, then ends `Cancelled` exit 0 with nothing installed. `stop_runtime_clients` (`install.sh:657`, called at `:712`) enumerates `ableton_prefix_pids`, names them, asks `i_q_stop_clients` (`ui.sh:316`, two lines above the new key at `:171`), refuses non-interactively without `--yes`, and calls `ableton_stop_prefix` (`lifecycle.sh:418`), which sends a prefix-scoped `wineserver -k` then re-checks runtime and prefix identity per pid before each signal. `lifecycle.sh:64-66` states the invariant the new gate breaks: "stop/report operations never act on a foreign Wine installation."
  ```diff
  -if { [ "$selected_action" = install ] || [ "$selected_action" = update ]; } \
  -   && pgrep -x wineserver >/dev/null 2>&1; then
  -    ui_question q_stop_wine_title n q_stop_wine_yes q_stop_wine_no
  -    if [ "$UI_ANSWER" != y ]; then
  -        cancelled=1; launch_hint=0; exit 0
  -    fi
  -    pkill -x wineserver >/dev/null 2>&1 || true
  -fi
  ```
  Fix: delete it, and the `q_stop_wine_*` entries with it.
- That one gate fails four suites, and CI cannot see it. On an idle box: `test-preflight-menu` 0 ok, aborting on its first assertion; `test-installer-lifecycle` 13 (was 81); `test-installer-ui` 20 (was 22); `test-release-policy` 6 (was 12). One defect under four names: `timed out waiting for @@DELEGATED@@`, `the delegated installer continues the header's step numbering`, `installer wrapper did not reach its child`, `stub-args: No such file or directory`. Three arms on one stub `.run` isolate it: as-is reaches no child, `pgrep` stubbed to find nothing reaches it with `args=install`, stubbed to always find one reaches no child. All four are in `make test`; run 33223004166 passed it at step 11 because a runner has no wineserver. Second time in three passes that a defect is green on the runner and red on a real machine, after the mawk simulator. Fix: start a throwaway wineserver before `make test` in `ci-pr-build.yml`.
- `nix/ableton-wine.nix:526`'s new `cmp -s` asserts every staged `scripts/lib/*.sh` is byte-identical to source, while `:291` runs `substituteInPlace` on `config.sh` in both staged copies and `:514` asserts that substitution happened. The builder fails `config.sh differs in libexec/lib`. `ci-flake` succeeded on main `0a61b5cd` (26 Aug) and failed on `2d9cf9e1`, `456b384c` and `bf7593d7`. Fix: skip `config.sh` in the `cmp` loop; `:514` already covers it.
- `uninstall.sh:481` makes a valid `$ABLETON_CONFIG_FILE` a hard precondition, so a missing or corrupt seven-line config strands the whole install. Reproduced on a real install by moving the config aside: `Step 3 Failed!`, 1 error, 1 second, with the runtime, launcher, 25 support files, 56 manifest rows and 7 desktop entries all left in place, and the footer sending the user to `Report this error` rather than naming the file to restore. Sandbox A/B across both heads, valid config as the control: `abedf95` removes everything whether the config is valid, deleted or corrupt; this head removes nothing and exits 1 in the latter two. The version replaced said the opposite in its own comment, "Installer settings are cleanup/retry convenience, not deletion authority". Fix: fall back to the runtime and state markers when the config is unusable.

Other comments:

- Ownership recording works, and it is unpinned. A real update rewrote the manifest to 56 rows with 44 of 46 file digests matching disk, against every changed file being stale before the fix. But `ABLETON_RECORD_PROJECT_FILES=1` at `install.sh:82` is referenced by no test: delete that line and every suite still passes while the manifest silently stops being written, which is the defect that took two passes and a live machine to find. Fix: assert a non-empty `install-manifest.tsv` after an install.
- The fix does not repair manifests the broken build corrupted. On the same machine the two remaining stale rows are `ableton-linkd` and `setup-link.sh`: Link is off, this run did not reinstall them, and `ableton_write_ownership_manifest` preserves rows for untouched paths including wrong ones. Uninstall will call exactly those "unrecognised or user-modified" and keep them. Fix: during install, reconcile rows whose digest does not match disk rather than carrying them forward.
- Both findings from the previous head are untouched. `setup-link.sh:647` is byte-identical across `108a0ae`, `abedf95` and `bf7593d`, so uninstall still reaches the overwrite question; `ui.sh` still takes no lock while `test-installer-ui.sh:333` asserts it must not, so two processes drawing the tree still collide. Fix: as previously described, no new information.
- `uninstall.sh` is on its third rewrite, 1837 to 546 lines, its suite 1194 to 467, executed assertions 38 to 22. Coverage came out of the file that then regressed above. Fix: no code change proposed; the next rewrite wants its assertions back first.
- `lib/preferences.sh` and the three preflight suites, 3782 lines, arrived mid-review and are implicated in none of the four blockers. I have not reviewed them. Fix: split into a separate PR against a merged 283.

Exercised in Live's own prefix: a full update completed 7/7 with 0 warnings and 0 errors, and the six new preflight questions render correctly. 243 assertions passed where 302 passed at `abedf95`, with three suites added. Next: `make test` and `ci-flake` green locally before the next push; four red suites and a red flake build are cheaper for you to clear than for me to re-derive.

